PR #6, opened 2020-09-23
Add basic version update script

eloquence:
## Description

Complements removal of version update logic in https://github.com/freedomofpress/securedrop/pull/5520 . In this context, it seemed most sensible to me to decouple, as it's consistent with other repos, and the docs release is one of the final steps in the release process.


zenmonkeykstop (approved):
LGTM, works as expected.

Issue #7, opened 2020-09-24
Update database diagram


eloquence:
We have a database diagram here:
https://github.com/freedomofpress/securedrop-docs/blob/main/docs/diagrams/securedrop-database.png

It is generated from:
https://github.com/freedomofpress/securedrop-docs/blob/main/docs/diagrams/securedrop-database.tex

It is outdated (last updated 2016) and misses schema changes (e.g., addition of UUIDs) and tables added since then. We should update it, and commit to keeping it up-to-date as part of every schema change.

eloquence:
(I've committed to work on this during the 10/1-10/15 sprint.)

PR #14, opened 2020-10-01
Update pull request template

joaedwar:
Pull requests should be opened against the `main` branch. For more information on contributing to SecureDrop documentation, see the [Documentation Guidelines](https://docs.securedrop.org/en/stable/development/documentation_guidelines.html).

## Status
*(What state is your PR in? Select one of the following and delete the option that does not apply.)*

Ready for review 

## Description of Changes

* Description: Comment out section prompts

* Fixes template

## Testing
*(How should the reviewer test this PR? Write out any special testing steps here.)*

## Release 
*(Any special considerations for release of this change into the stable version of the documentation?)*

* n/a

## Checklist (Optional)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,7 +2,7 @@ Pull requests should be opened against the `main` branch. For more information o
 
 
 ## Status
-*(What state is your PR in? Select one of the following and delete the option that does not apply.)*
+<!-- What state is your PR in? Select one of the following and delete the option that does not apply. -->
 
 Ready for review / Work in progress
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -11,16 +11,15 @@ Ready for review / Work in progress
 
 * Description: 
 
-* Fixes Issue#<issue-number>
+* Fixes [Issue#<issue-number>](url)
```
eloquence, line 14: Let's keep this simple, no need to add markdown formatting. In fact, note that GitHub auto-closes issues once the associated PR is merged if it follows the format:

`Fixes #123` (where 123 is the issue number)

So I would change this in the template to:

`Fixes Issue#`

eloquence (changes requested):
Thanks @joaedwar, using HTML comments makes sense. How about also wrapping the intro ("Pull requests should be opened against the `main` branch. ...") in a comment? Alternatively, to be honest, I think we could do away with that intro altogether. We can use the file `CONTRIBUTING.md` in the root to point people to the documentation guidelines, and `main` is the default target, so I don't think it needs spelling out on every PR.

One other review comment inline.

joaedwar:
@eloquence  is this okay to merge?

eloquence:
Yep, let's take it for a spin. Let me quickly squash and rebase it.

eloquence:
Unfortunately I can't squash it for you:
> batch response: You need Push access to upload Git LFS objects

This is a side effect of us using LFS for this repo. Because the rebase includes the LFS commits from #7, and this is a contributor fork, I can't push up those commits, and the rebase fails. To ease collaboration on future collaboration, it would be best if you added me as a collaborator to your fork, which will give me the ability to push.

In the meantime, if you have time, you can also go ahead and rebase/squash, and then we can merge.

PR #25, opened 2020-10-09
Fix YubiKey link, ignore-list Weblate URL

eloquence:
Makes link check happy for a few more hours/days of its frequently unhappy existence. The Weblate URL is ignore-listed because it requires authentication.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -322,6 +322,7 @@ linkcheck_ignore = [
     "https://forum.securedrop.org/admin/users/list/active",
     "https://weblate.securedrop.org/projects/securedrop/securedrop/#repository",
     "https://github.com/freedomofpress/securedrop-debian-packages-lfs",
+    r"https://weblate.securedrop.org/accounts/profile/.*",
     r"https://github.com/freedomofpress/securedrop/issues/.*",
     r"https://github.com/freedomofpress/securedrop/tree/.*",
 ]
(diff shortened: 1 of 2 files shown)


rocodes (commented):
When I check this branch out and run `make docs-linkcheck`, I get:

```
Warning, treated as error:
/home/user/projects/securedrop-docs/docs/development/tips_and_tricks.rst.rst:134:broken link: https://github.com/freedomofpress/securedrop-docs/tree/main/docs/diagrams (429 Client Error: too many requests for url: https://github.com/freedomofpress/securedrop-docs/tree/main/docs/diagrams)
make: *** [Makefile:17: docs-linkcheck] Error 2
```

eloquence:
@rocodes This is due to linkcheck hitting GitHub's rate limit. See previous issues with GitHub URLs documented in https://github.com/freedomofpress/securedrop/pull/5396 . I only get the 429 sporadically. We do have the worker increase in our `conf.py`; @rmol may have ideas for other things to try. This is a URL we _do_ want to monitor if at all possible.

I would suggest we merge this if the URL itself looks fine, and addressing further improvements in follow-up.

rocodes (approved):

rocodes:
Sounds good @eloquence. Approved + merging

rmol:
We could submit a PR to teach linkcheck to be smarter about rate-limited domains. That would not help with the issue at hand for a while. 

We could complicate our `docs-linkcheck` target with a script that runs linkcheck, collects the errors from the output and retries any 429s, with delays, until they're all checked, and only if we exceed a reasonable amount of time or number of retries, or still have other types of errors, exit non-zero.

PR #71, opened 2020-10-21
Modify v3 migration instructions to emphasize concurrent upgrade path

rocodes:
## Status
Ready for review

## Description of Changes

* Simplify explanation of onion services as much as possible  
* Update v3 migration docs to emphasize concurrent v2+v3 migration path
* Update information on v2 support timeline in SecureDrop
* Clearly enumerate upgrade steps  
* Add note about human-readable onion URLs and link to blog post

## Testing
* visual review + linting

## Release 
* n/a 

## Checklist (Optional)

### DIFF
--- a/docs/v3_services.rst
+++ b/docs/v3_services.rst
@@ -1,89 +1,106 @@
 SecureDrop V3 Onion Services
 ============================
 Tor onion services provide anonymous inbound connections to websites and other
-servers exclusively over the Tor network. For example, SecureDrop uses onion services
-for the *Journalist* and *Source Interface* websites, as well as for
-adminstrative access to the servers in SSH-over-Tor mode.
+servers exclusively over the Tor network. SecureDrop uses onion services 
+for the *Journalist* and *Source Interface* websites, as well as for 
+adminstrative access to the servers in SSH-over-Tor mode. 
 
-SecureDrop currently supports both the older v2 version of the onion services
-protocol, and the current version, v3. The current version provides stronger
-cryptographic primitives than v2 onion services, and includes redesigned
-protocols that guard against service information leaks on the Tor network.
+Currently, SecureDrop supports two versions of the onion services
+protocol: the older version (v2), and the current version (v3), which provides 
+stronger cryptography and includes redesigned protocols that guard against 
+service information leaks on the Tor network. Because of these important 
+improvements, the Tor project is
+`deprecating support for v2 onion services <https://blog.torproject.org/v2-deprecation-timeline>`__. 
 
-Because of these important improvements, the Tor project is
-`deprecating support for v2 onion services <https://blog.torproject.org/v2-deprecation-timeline>`__.
-SecureDrop will remove support for v2 onion services as part of its 2.0.0
-release, planned for **February 2021**. If you are currently using v2 onion services,
-they will become unreachable at that point.
+SecureDrop will begin phasing out support for v2 onion services in 
+**February 2021**. All newsrooms should enable v3 onion services as soon as
+possible, but may continue using v2 concurrently to allow all journalists and 
+sources to transition smoothly.
 
-The unique identifier in v3 onion addresses is 56 characters long - for example:
+.. warning::
+
+   If you are still using 16-character v2 onion URLs in April 2021, your SecureDrop 
+   servers will become unreachable, and you will have to reinstall SecureDrop 
+   from scratch.
+
+How to distinguish between v2 and v3 onion services
+---------------------------------------------------
+
+If your *Source Interface* address is 56 characters long, you are already using v3 onion services, as in the following example:
 
 .. code-block:: none
 
-  http://vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd.onion/
+  http://sdolvtfhatvsysc6l34d65ymdwxcujausv7k5jk4cy5ttzhjoi6fzvyd.onion/
 
-This makes them easily distinguishable from 16-character v2 onion addresses,
-such as:
+If your *Source Interface* address is 16-characters long, you are still using 
+v2 onion services, and need to transition to v3 in order to keep using 
+SecureDrop, as in the following example:
 
 .. code-block:: none
 
   http://secrdrop5wyphb5x.onion/
 
-.. important::
+What an upgrade entails
+-----------------------
 
-   If you are currently advertising a 16-character v2 address like the above
-   to your sources, it will become unreachable in February 2021. You must
-   update to v3 onion services before then.
+Enabling v3 onion services will have the following effects:
 
-   There is no downgrade path from v3 to v2 using the ``securedrop-admin``
-   tool. We recommend that you follow the v2+v3 migration path below, and test v3
-   functionality thoroughly before :ref:`disabling v2 onion services <disable_v2>`.
-
-Migrating from v2 to v3 only or v2+v3
--------------------------------------
-
-Preparing for the migration
-^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


docs/v3_services.rst
```diff
@@ -1,89 +1,103 @@
 SecureDrop V3 Onion Services
 ============================
-Tor onion services provide anonymous inbound connections to websites and other
```
zenmonkeykstop, line 3: It's worth keeping the "what is an onion service and what does it do?" line IMO, just for admins and other folk that don't spend all their time on the DarkWeb, and may not immediately twig why this is important.

docs/v3_services.rst
```diff
@@ -1,89 +1,103 @@
...
+  your landing page and other resources that reference your *Source Interface* 
+  address, such as your SecureDrop directory entry.
+- *Journalist* and *Admin Workstations* will need to be updated by your 
+  administrator so that they contain the new onion service addresses and
+  credentials.
 - If your instance uses HTTPS, you will need to provision a new certificate for
   the v3 *Source Interface* address - this will need to be done `after` the new
   address has been generated.
 
-.. note:: If your certificate provisioning process requires validation of the
```
zenmonkeykstop, line 64: Again, this is overly simple IMO, we should be upfront with folks about why they need to contact us.

docs/v3_services.rst
```diff
@@ -1,89 +1,103 @@
...
-          new v3 domain, you may not be able to complete the v3 migration process
-          without first disabling HTTPS for v2. If your instance currently uses
+.. note:: If your instance currently uses
           HTTPS with an EV certificate, please contact us via the `SecureDrop
-          support portal`_ or via email to securedrop@freedom.press
-          before proceeding with the migration. Please use `our GPG key`_ for
-          any email communication.
+          support portal`_ or via `GPG-encrypted`_ email to securedrop@freedom.press
```
zenmonkeykstop, line 61: Pet peeve: where possible, link text should indicate what the link is pointing to. "our GPG key" is pretty unambiguous. "Gpg-encrypted" is an adjective. Is it a link to something describing GPG? is it a link to a key? it's a mystery!

docs/v3_services.rst
```diff
@@ -94,38 +108,45 @@ Workstation* with ``./securedrop-admin tailsconfig``.
 
     cd ~/Persistent/securedrop
 
-- Verify that SecureDrop version 1.0.0 or greater is available or installed on
+- Verify that SecureDrop version |version| is available or installed on
```
zenmonkeykstop, line 111: Good use of `version` to remove ambiguity in original doc. :white_check_mark:

zenmonkeykstop (changes requested):

docs/v3_services.rst
rocodes: IMO our docs have too much such preamble and my bias would be to keep them to the requirements ("enable v3 today! here's how") instead of taking as long as we do to explain why; however, maybe this is more of a long-term structural question about our documentation and should be discussed separately. Added this back in 71d4739.

docs/v3_services.rst
rocodes: I just switched the order of these sentences, nothing was removed. Do you still feel like it's too simplistic? If so I can revert, but I think the note/callout reads clearer this way.

docs/v3_services.rst
rocodes: fixed in 71d4739

docs/v3_services.rst
eloquence: tiny typo slipped in: TOr->Tor

docs/v3_services.rst
```diff
@@ -1,89 +1,105 @@
...
+
+How to distinguish between v2 and v3 onion services
+---------------------------------------------------
+
+If your *Source Interface* address is 56 characters long, you are already using v3 
+onion services:
 
 .. code-block:: none
 
   http://vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd.onion/
```
eloquence, line 34: We can now replace that with `sdolvtfhatvsysc6l34d65ymdwxcujausv7k5jk4cy5ttzhjoi6fzvyd.onion`, the SecureDrop.org v3 address.

docs/v3_services.rst
```diff
@@ -1,89 +1,105 @@
...
 
-This makes them easily distinguishable from 16-character v2 onion addresses,
-such as:
+If your *Source Interface* address is 16-characters long, you are still using 
+v2 onion services, and need to transition to v3 in order to keep using SecureDrop:
 
 .. code-block:: none
 
-  http://secrdrop5wyphb5x.onion/
+  http://secrdrop5wyphb5x.onion/
```
eloquence, line 41: Note that this address is now dead (it was the pre-v2 SecureDrop.org address).

docs/v3_services.rst
```diff
@@ -1,89 +1,105 @@
...
+How to upgrade
+--------------
+
+   #. :ref:`Prepare backups <prepare_backups_v3>`
+   #. :ref:`Enable v3 onion services <enable_v3>` alongside v2, to avoid 
+      disruption to journalists and sources 
+   #. :ref:`(Optional) Enable HTTPS <enable_https_v3>`
+   #. :ref:`Update Tails workstations <update_tails_v3>` with the new v3 onion addresses
+   #. :ref:`Publish <publish_v3>` your new *Source Interface* URL      
+   #. Once you are satisfied with these changes, :ref:`disable v2 onion services<disable_v2>`.
```
eloquence, line 82: This is a very nice overview! :-)

docs/v3_services.rst
```diff
@@ -347,3 +376,14 @@ When you're ready, follow the steps below to transition to v3 services only:
 
     ./securedrop-admin sdconfig   # choose "no" for v2, "yes" for v3
     ./securedrop-admin tailsconfig
+
+
+Human-readable onion URLs
```
eloquence, line 381: Thanks for adding this section! The feature name (as also referenced in the Tor browser UI if you click on extended information for an onion address) is "onion name", which I think we should use at least once in this paragraph.

Since it's likely to come up again, what do you think about maybe explicitly adding a sentence here about "vanity addresses", e.g.:

> While it is possible to generate "vanity addresses" that start with a few recognizable letters, we are unable to provide technical assistance with any such process, and we recommend registering an onion name instead.

eloquence:
Thanks @rocodes, I quite like the improved organization of the procedural steps; once comments are resolved, I think this is good to go.

docs/v3_services.rst
rocodes: Added 'onion names' terminology in 61fa02a, and also added

> Onion names are the supported solution for organizations wishing to 
> customize their *Source Interface* URL, and are recommended over "vanity"
> URLs (addresses that start with a few recognizable characters).

docs/v3_services.rst
rocodes: Added the "as in the following example" back in in 61fa02a in both cases, and the updated v3 URL.

docs/v3_services.rst
rocodes: It is. But I didn't know what to replace it with, since theoretically all v2s will someday be deprecated (and I didn't want to call out anyone specifically by saying "hey look at _these guys_ still using v2...")  I have left it as is for now, and I think it's clear that the example is just an example.  If strong feelings about this, I will add a ` # just an example of v2 formatting` type comment beside the URL.

docs/v3_services.rst
rocodes: Oops- fixed in 61fa02a, thanks for catch

eloquence (approved):
LGTM, thanks @rocodes!

PR #97, opened 2020-10-29
Update average throughput onion

KwadroNaut:
Checked the data available in metrics.torproject.org which confirmed my suspicion: it's faster as it once was. My eyeball guess is that a the average throughput to onions is rather 2.5Mbps, conservative. 300 kB/s is even less, but double the speed as what was written here.

## Status

Ready for review 

## Description of Changes

* Description: update on Tip for admins regarding average throughput of onions.

## Testing

* see metrics.torproject.org

### DIFF
--- a/docs/backup_and_restore.rst
+++ b/docs/backup_and_restore.rst
@@ -25,7 +25,7 @@ encourage Journalists to delete regularly unneeded submissions from the *Journal
 Interface*.
 
 .. tip:: Although it varies, the average throughput of an onion service is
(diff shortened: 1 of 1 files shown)


rocodes (commented):
Looking at [stats from 2020](https://metrics.torproject.org/onionperf-throughput.html?start=2020-01-31&end=2020-10-29&server=onion) the median avg throughput (excluding zero values for unavailable nodes) is 2.45 Mbps, so good eyeballing, @KwadroNaut.  This LGTM.

eloquence (approved):
If anything this is still pretty conservative as you say, so LGTM as well.

PR #99, opened 2020-11-02
Update Admin Workstation Setup doc to resolve tag/branch ambiguity

### LINKED ISSUE #42: Admin Workstation setup docs should warn about tag/branch name ambiguity
In workstation update docs, we've pretty consistently used the following sequence for code updates of the SecureDrop repository:

1) Fetch tags
2) Get key
3) Verify tag
4) Checkout tag
5) Ensure you don't get the "refname $foo is ambiguous" warning

([Example](https://docs.securedrop.org/en/stable/upgrade/1.5.0_to_1.6.0.html#performing-a-manual-update))

The [Admin Workstation setup docs](https://docs.securedrop.org/en/release-0.13.0/set_up_admin_tails.html#download-the-securedrop-repository) use this sequence:

1) Get key
2) Checkout tag
3) Verify tag

No warning about branch/tag ambiguity (if there's a `0.13.0` branch, it will happily be checked out, even if it's not signed).

We should follow the updated standard sequence here as well:

1) Get key
2) Verify tag
3) Checkout tag
4) Ensure you don't get the "refname $foo is ambiguous" warning

joaedwar:
## Status

Ready for review 

## Description of Changes

* Description: Update the Admin Workstation setup docs to standardize the sequence.

* Fixes #42 

## Testing

*

## Release 

* 

## Checklist (Optional)

### DIFF
--- a/docs/set_up_admin_tails.rst
+++ b/docs/set_up_admin_tails.rst
@@ -136,17 +136,33 @@ signed with the release signing key:
 .. code:: sh
 
     cd ~/Persistent/securedrop/
-    git checkout 1.6.0
+    git fetch --tags
+    gpg --keyserver hkps://keys.openpgp.org --recv-key \ "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"
     git tag -v 1.6.0
 
-You should see ``Good signature from "SecureDrop Release Signing Key"`` in the
-output of that last command along with the fingerprint above.
+The output should include the following two lines:
+
+.. code:: sh
+
+    gpg:                using RSA key 22245C81E3BAEB4138B36061310F561200F4AD77
+    gpg: Good signature from "SecureDrop Release Signing Key"
+
 
 .. important::
 
    If you do not see the message above, signature verification has failed
    and you should **not** proceed with the installation. If this happens,
    please contact us at securedrop@freedom.press.
+
+Verify that each character of the fingerprint matches what is on the 
+screen of your workstation. If it does, you can check out the new release:
+
+.. code:: sh
+
+    git checkout 1.6.0
+
+.. include:: ../includes/refname-warning.txt
+
 
 .. _keepassxc_setup:
 
--- a/docs/set_up_admin_tails.rst
+++ b/docs/set_up_admin_tails.rst
@@ -161,7 +161,7 @@ screen of your workstation. If it does, you can check out the new release:
 
     git checkout 1.6.0
 
-.. include:: ../includes/refname-warning.txt
+.. include:: includes/refname-warning.txt
 
 
 .. _keepassxc_setup:
--- a/docs/deployment/landing_page.rst
+++ b/docs/deployment/landing_page.rst
@@ -182,9 +182,9 @@ If you intend to run nginx as your webserver instead, this will work:
     add_header X-Frame-Options DENY;
     add_header X-XSS-Protection "1; mode=block";
     add_header X-Content-Type-Options nosniff;
-    add_header X-Download-Options: noopen;
+    add_header X-Download-Options noopen;
     add_header X-Permitted-Cross-Domain-Policies master-only;
-    add_header Content-Security-Policy "default-src 'none'; script-src 'self'; style-src 'self'; img-src 'self'; font-src 'self';"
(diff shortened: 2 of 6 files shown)


docs/set_up_admin_tails.rst
```diff
@@ -136,18 +136,37 @@ signed with the release signing key:
...
     cd ~/Persistent/securedrop/
-    git checkout 1.6.0
+    git fetch --tags
+    gpg --keyserver hkps://keys.openpgp.org --recv-key \
+    "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"
     git tag -v 1.6.0
 
 You should see ``Good signature from "SecureDrop Release Signing Key"`` in the
-output of that last command along with the fingerprint above.
+output.
```
eloquence, line 145: This paragraph is duplicated by the new, improved language below, so you can remove it.

docs/set_up_admin_tails.rst
```diff
@@ -136,18 +136,37 @@ signed with the release signing key:
...
    If you do not see the message above, signature verification has failed
    and you should **not** proceed with the installation. If this happens,
    please contact us at securedrop@freedom.press.
 
+Verify that each character of the fingerprint matches what is on the 
+screen of your workstation. If it does, you can check out the new release:
+
+.. code:: sh
+
+    git checkout 1.6.0
```
eloquence, line 166: After this step, we want to insert the language we've been using in the upgrade guides, i.e.

```
.. important:: If you do see the warning "refname '1.6.0' is ambiguous" in the
  output, we recommend that you contact us immediately at securedrop@freedom.press
  (`GPG encrypted <https://securedrop.org/sites/default/files/fpf-email.asc>`__).
```

We could move this into an include to reduce the amount of copy/paste.

docs/set_up_admin_tails.rst
```diff
@@ -136,18 +136,37 @@ signed with the release signing key:
 .. code:: sh
 
     cd ~/Persistent/securedrop/
-    git checkout 1.6.0
+    git fetch --tags
+    gpg --keyserver hkps://keys.openpgp.org --recv-key \
+    "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"
```
eloquence, line 141: To indicate that this is a continuation from the previous line, I would indent it with a single space.

eloquence (changes requested):
Thanks @joaedwar, this one is almost ready, see comments :)

docs/set_up_admin_tails.rst
joaedwar: Just out of curiosity, what does this warning suggest? Is it a security issue? Why should a user contact immediately?

conorsch: We assume that `1.6.0` would refer to a signed git tag on the repository at https://github.com/freedomofpress/securedrop . When running `git checkout 1.6.0`, however, git will also happily checkout a branch, rather than a tag, if the branch is named "1.6.0". In that case, it's not clear to git which the user wanted. We always want the checkout to result in a tag, but there's no flag we can pass to git to make a branch name unacceptable as a refspec for the checkout command.

eloquence: (Just a note that the include `refname-warning.txt` is still missing from this PR, in case you forgot to push a commit.)

docs/set_up_admin_tails.rst
```diff
@@ -136,18 +136,34 @@ signed with the release signing key:
...
    please contact us at securedrop@freedom.press.
 
+Verify that each character of the fingerprint matches what is on the 
+screen of your workstation. If it does, you can check out the new release:
+
+.. code:: sh
+
+    git checkout 1.6.0
+
+.. include:: ../includes/refname-warning.txt
```
eloquence, line 164: This include will fail because `..` specifies the parent directory, but the included file is under `docs/includes` and the doc you're modifying is under `docs`. You can see the failure in the lint error output on this PR: https://app.circleci.com/pipelines/github/freedomofpress/securedrop-docs/108/workflows/42ed5dc1-fe37-4fb7-9eaa-14a28fed3db5/jobs/104

To correct it, specify `includes/refname-warning.txt`

PR #100, opened 2020-11-02
Replace instances for OSSEC Alert Key

### LINKED ISSUE #44: Define and standardize OSSEC Alert Key terminology
Similar to freedomofpress/securedrop#4122, let's standardize how we describe the keypair used for OSSEC alerts, and describe this in the glossary. For example, we sometimes call it "the GPG key that OSSEC will encrypt alerts to", but in the installation guide we also refer to it as "the admin’s GPG public key file (for encrypting OSSEC alerts)". Repeating the same long definition over and over again risks introducing small inconsistencies like that.

I would recommend that we call it the *OSSEC Alert Key* or *OSSEC Alert Public Key*, and that we call the key for the 24h alerts the *Journalist Alert Key* or *Journalist Alert Public Key* (use the most precise qualifier -- public or private -- where possible).

joaedwar:
## Status

Ready for Review

## Description of Changes

* **Description**: Define and standardize OSSEC Alert Key terminology. Replace "the admin’s GPG public key file (for encrypting OSSEC alerts)" and similar description with OSSEC Alert Key or OSSEC Alert Public Key, and call the key for the 24h alerts the Journalist Alert Key or Journalist Alert Public Key.

* Fixes #44

## Testing

*

## Release 

* 

## Checklist (Optional)

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -49,13 +49,12 @@ Instructions for using SecureDrop as a *Journalist* are available in our
 :doc:`Journalist Guide <journalist>`.
 
 
-Journalist Alert Key
---------------------
-The *Journalist Alert Key* is used for encrypting the alert that notifies journalists daily 
-via encrypted email about whether or not there has been submission activity 
-in the past 24 hours. More information about journalist alerts, is in our 
-:doc:`Journalist Guide <journalist>`.
-
+Journalist Alert Public Key
+---------------------------
+The *Journalist Alert Public Key* is used for encrypting the :ref:`daily alert <daily_journalist_alerts>`
+that notifies journalists via encrypted email about whether or not there has been
+submission activity in the past 24 hours. The journalist uses an associated
+private key to decrypt the alerts.
 
 Journalist Interface
 --------------------
@@ -98,8 +97,9 @@ to this server, and they may only do so using Tor.
 OSSEC Alert Public Key
 ----------------------
 The *OSSEC Alert Public Key* is the GPG key that OSSEC will encrypt alerts to.
-It is used by the admin to access encrypted OSSEC alerts from the Monitor Server.
-Instructions about setting up OSSEC alerts are available in the :doc:`OSSEC Guide <ossec_alerts>`.
+The associated private key is used by the admin to access encrypted OSSEC alerts
+from the *Monitor Server*. Instructions for setting up OSSEC alerts can be found
+in the :doc:`OSSEC Guide <ossec_alerts>`.
 
 
 .. _svs:
@@ -133,7 +133,7 @@ SecureDrop to communicate with a *Journalist*. A *Source* will always
 access SecureDrop through the *Source Interface* and must do so using Tor.
 
 Instructions for using SecureDrop as a *Source* are available in our
-:doc: `Source Guide <source>`.
+:doc:`Source Guide <source>`.
 
 
 Source Interface
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -6,19 +6,36 @@ diagram <https://docs.securedrop.org/en/latest/overview.html#infrastructure>`__,
 are specific to SecureDrop. The list below attempts to enumerate and
 define these terms.
 
-Source
-------
 
-The *Source* is the person who submits documents to SecureDrop and may use
-SecureDrop to communicate with a *Journalist*. A *Source* will always
-access SecureDrop through the *Source Interface* and must do so using Tor.
+Admin Workstation
+-----------------
+The *Admin Workstation* is a machine that the system admin can
+use to connect to the *Application Server* and the *Monitor Server* using Tor
+and SSH. The admin will also need to have an Android or iOS
+device with the FreeOTP app installed.
 
-Instructions for using SecureDrop as a *Source* are available in our
-`Source Guide <https://docs.securedrop.org/en/latest/source.html>`__.
+
+Application Server
+------------------
+The *Application Server* runs the SecureDrop application. This server hosts both
+the website that sources access (the *Source Interface*) and the website that
+journalists access (the *Journalist Interface*). Sources, journalists, and
+admins may only connect to this server using Tor.
+
+
+Export Device
+-------------
+The *Export Device* is the physical media (e.g., designated USB drive) used to
+transfer decrypted documents from the *Secure Viewing Station* to a journalist's
+everyday workstation, or to another computer for additional processing.
+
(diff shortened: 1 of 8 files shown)


docs/install.rst
```diff
@@ -75,7 +75,7 @@ continuing:
 -  the *Submission Public Key* (from the *Transfer Device*)
 -  the *Submission Key* fingerprint
 -  the email address that will receive alerts from OSSEC
--  the GPG public key and fingerprint for the email address that will
+-  the OSSEC Alert Public Key and fingerprint for the email address that will
```
eloquence, line 78: Per our standard convention, we should italicize this term and add a definition to the glossary (happy to talk through what this key is used for).

eloquence (changes requested):
Thanks @joaedwar! See inline comment; the biggest requested change is to add a glossary entry. Note that you currently have a lint failure.

joaedwar:
@eloquence

docs/glossary.rst
```diff
@@ -31,41 +48,17 @@ being transferred to an Internet-connected computer.
...
 
-Source Interface
-----------------
-
-The *Source Interface* is the website that sources will access to
-submit documents and communicate with journalists. This site is
-hosted on the *Application Server* and can only be accessed through Tor.
+Journalist Alert Key
+--------------------
+The *Journalist Alert Key* is an alert that notifies journalists daily
```
eloquence, line 54: is an alert -> is used for encrypting the alert

docs/glossary.rst
```diff
@@ -31,41 +48,17 @@ being transferred to an Internet-connected computer.
...
-
-The *Source Interface* is the website that sources will access to
-submit documents and communicate with journalists. This site is
-hosted on the *Application Server* and can only be accessed through Tor.
+Journalist Alert Key
+--------------------
+The *Journalist Alert Key* is an alert that notifies journalists daily 
+via encrypted email about whether or not there has been submission activity 
+in the past 24 hours. More information about journalist alerts, is in our 
+`Journalist Guide <https://docs.securedrop.org/en/stable/journalist.html#daily-journalist-alerts-about-submissions>`__.
```
eloquence, line 57: We should not use absolute URLs to link to other chapters of the documentation. reStructured text supports the `:doc:` and `:ref:` syntax to cross-reference other chapters and sections, respectively. The glossary chapter has other absolute URL references which should be replaced with `:doc`: or `:ref:`. If you want to take a stab at cleaning up the links in this chapter, please add a commit to do so, otherwise I can do it before merging.

docs/glossary.rst
```diff
@@ -31,41 +48,17 @@ being transferred to an Internet-connected computer.
...
-page is hosted as a standard (i.e. non-Tor) webpage on the news organization's
-site. It provides first instructions for potential sources.
 
-Source Interface
-----------------
-
-The *Source Interface* is the website that sources will access to
-submit documents and communicate with journalists. This site is
-hosted on the *Application Server* and can only be accessed through Tor.
+Journalist Alert Key
```
eloquence, line 52: If we introduce this term newly to the glossary, we should make sure we do in fact use it in the documentation, i.e. the chapter about journalist alerts: https://docs.securedrop.org/en/stable/journalist.html#daily-journalist-alerts-about-submissions

docs/threat_model/mitigations.rst
```diff
@@ -182,15 +182,17 @@ Attacks on Network Infrastructure
...
+   Freedom.press, torproject.org Tor activity, SD submissions may be correlated
 -  SecureDrop.org is compromised
 -  User web traffic to SecureDrop Landing Page uses CDN and may be logged
 -  Tor network exploit
 -  apt server man-in-the-middle used to serve old or malicious packages
 -  SecureDrop apt servers are compromised, or apt server man-in-the middle attack injects malicious packages
 -  News Organization network is compromised
--  OSSEC and/or Journalist alert SMTP account credentials compromised
--  OSSEC and/or Journalist alert private key compromised
+-  *OSSEC Alert Public Key* and/or *Journalist Alert Public Key* SMTP account
```
eloquence, line 193: This line refers to the email account credentials used to receive emails, not the keys used to encrypt them. So the original language is correct here, and should not be modified.

docs/threat_model/mitigations.rst
```diff
@@ -182,15 +182,17 @@ Attacks on Network Infrastructure
...
 -  User web traffic to SecureDrop Landing Page uses CDN and may be logged
 -  Tor network exploit
 -  apt server man-in-the-middle used to serve old or malicious packages
 -  SecureDrop apt servers are compromised, or apt server man-in-the middle attack injects malicious packages
 -  News Organization network is compromised
--  OSSEC and/or Journalist alert SMTP account credentials compromised
--  OSSEC and/or Journalist alert private key compromised
+-  *OSSEC Alert Public Key* and/or *Journalist Alert Public Key* SMTP account 
+   credentials are compromised
+-  *OSSEC Alert Public Key* and/or *Journalist Alert Private Key* are compromised
```
eloquence, line 195: In both cases the *private* key is what we need to worry about most; the *public* key will not allow an adversary to read encrypted emails.

docs/set_up_admin_tails.rst
```diff
@@ -204,7 +204,7 @@ the template are:
 - Email account for sending OSSEC alerts
 - Monitor Server SSH Onion URL
 - Network Firewall Admin Credentials
-- OSSEC GPG Key
+- OSSEC Alert Public Key
```
eloquence, line 207: Let's italicize here as well.

docs/glossary.rst
```diff
@@ -87,19 +80,39 @@ documents, and move them to the *Secure Viewing Station* using the
...
+email alert if something seems wrong. Only system admins connect
+to this server, and they may only do so using Tor.
+
+
+OSSEC Alert Public Key
+----------------------
+The *OSSEC Alert Public Key* is the GPG key that OSSEC will encrypt alerts to.
+It is used by the admin to access encrypted OSSEC alerts from the Monitor Server.
+
+Copy the *OSSEC Alert Public Key* to ``install_files/ansible-base`` and then
```
eloquence, line 103: These instructions do not belong in the glossary.

eloquence (changes requested):
Thanks for your continued work on this, @joaedwar. Still a bit more cleanup required, happy to go over this in more detail in our 1:1 :)

eloquence (approved):
Thanks @joaedwar, fixes look good! Rebased & made some final minor tweaks, approving with those changes.

PR #101, opened 2020-11-06
clarified storage option for nuc8i5bek

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Minor fix, clarifies that the NUC8i5BEK is M.2-only.


rocodes (approved):
Thank you @zenmonkeykstop!

PR #115, opened 2020-12-10
Standardize on 'i18n' label for i18n-related GitHub issues

eloquence:
We were using both the `goals:` variant and the one without prefix
before.

Also updates other label links that were yielding zero results.

## Status

Ready for review

## List of checkboxes 


creviera (approved):
new links work and makes sense to remove links to issues with labels that no longer exist

PR #120, opened 2021-01-03
Add clarification on translation of HTML content

### LINKED ISSUE #83: Translator documentation needs clarification around HTML content
## Description

We're seeing some HTML attribute content being translated that shouldn't be, and upon reviewing [our docs about that](https://docs.securedrop.org/en/latest/development/l10n.html#html-content) I believe we could be more prescriptive about which attributes should be translated -- pretty much _just_ `alt` or `title` -- and we should also enumerate attributes which should _never_ be translated: `class`, `id`, `href`, `src`, and so on.

## User Research Evidence

The documentation is clearly not explicit enough, as translators don't have surety regarding the translation of HTML content.

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

Clarifies that both `alt` and `title` HTML attributes must be translated, and no other attribute should. Add example to demonstrate correct translations for both the `alt` and `title` tags in plausible contexts.

Also add a list of currently used HTML attributes which values should not be translated for explicitness. I tried to make clear that the list is not meant to be exhaustive, but listed all attributes currently in use in the [SecureDrop Core translations](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/translations/messages.pot).

* Fixes #83

## Testing

- [ ] Please proof-read, I'm not an native English speaker :slightly_smiling_face: and don't hesitate to amend or ask me to correct where needed!

## Release 

I don't think any special considerations are needed before releasing this change.

## Checklist (Optional)

### DIFF
--- a/docs/development/l10n.rst
+++ b/docs/development/l10n.rst
@@ -159,15 +159,22 @@ HTML content
 ^^^^^^^^^^^^
 
 Some source strings represent HTML that will be presented in the
-SecureDrop web interface. It can be hard to tell what to translate,
-since HTML is source code and changing the wrong thing can break the
-page layout.
+SecureDrop web interface.
 
-One thing you should always translate are ``alt`` attributes of image
-elements. Image elements (``<img>``) in HTML place a picture on the
-page. People with visual impairments rely on a special note on the
-image element -- the ``alt`` attribute -- to describe the image, so
-it's helpful to translate those. Here's an example that contains an
+HTML elements (embraced by in ``<``, ``>``, example: ``<strong>``)
+can contain multiple so-called *attributes*.
+
+The text of the two attributes called ``alt`` and ``title``
+should be translated. The text of the other attributes should not
+be translated.
+
+Attribute ``alt``
+"""""""""""""""""
+
+Image elements (``<img>``) in HTML place a picture on the
+page. Because people with visual impairments rely on a special note
+on the image element -- the ``alt`` attribute -- to describe the image,
+it is necessary to translate those. Here's an example that contains an
 image with both an ``alt`` attribute *and* a placeholder::
 
   <img src="{icon}" alt="shield icon">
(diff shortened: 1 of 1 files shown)


gonzalo-bulnes:
**Question**: Are the SecureDrop Client strings in the same `.pot` as the strings from SecureDrop Core?

eloquence:
(Rebased.)

eloquence:
> Question: Are the SecureDrop Client strings in the same .pot as the strings from SecureDrop Core?

The SecureDrop Client is not yet translatable. :( This is tracked in https://github.com/freedomofpress/securedrop-client/issues/239

docs/development/l10n.rst
```diff
@@ -159,15 +159,22 @@ HTML content
...
+
+The text of the two attributes called ``alt`` and ``title``
+should be translated. The text of the other attributes should not
+be translated.
+
+Attribute ``alt``
+"""""""""""""""""
+
+Image elements (``<img>``) in HTML place a picture on the
 page. People with visual impairments rely on a special note on the
```
eloquence, line 175: I would write "Because people with visual impairments rely ..., it is necessary ...".

Right now you say "People with visual impairments rely ..., it is necessary ..." which sounds a bit odd in English because the two clauses of the sentence are not connected.

docs/development/l10n.rst
```diff
@@ -179,6 +186,36 @@ translated HTML in Portuguese would be::
...
+should be translated.
+
+Other attributes
+""""""""""""""""
+
+No attribute other than ``alt`` and ``title`` should be translated.
+
+In particular, please make sure the attributes ``class``, ``id``,
+``height``, ``href``, ``rel``, ``src`` and ``width``
+are never translated.
```
eloquence, line 217: I defer to @rmol here, but it might be worth qualifying this a little bit for `href` -- there may be situations where linking to a different URL in the target language would be appropriate. Perhaps simply asking translators to comment in Weblate when in doubt is the way to go here.

docs/development/l10n.rst
```diff
@@ -159,15 +159,22 @@ HTML content
...
-since HTML is source code and changing the wrong thing can break the
-page layout.
+SecureDrop web interface.
 
-One thing you should always translate are ``alt`` attributes of image
-elements. Image elements (``<img>``) in HTML place a picture on the
+HTML elements (embraced by in ``<``, ``>``, example: ``<strong>``)
+can contain multiple so-called *attributes*.
+
+The text of the two attributes called ``alt`` and ``title``
```
eloquence, line 167: As a general note, I would suggest changing phrases like  "the text of the attribute X" to "the text content of the attribute X", to make it even clearer that we are not talking about the attribute name itself.

eloquence (changes requested):
The additional explanatory text is really great, thanks @gonzalo-bulnes. Some nits in the review, and would appreciate a quick visual review from @rmol or @kushaldas on this one.

docs/development/l10n.rst
gonzalo-bulnes: Very good point. That will also be highly dependent on the URL.

_Thinking aloud: in a case like this, could the localized URLs be given to the translators via string context?_

gonzalo-bulnes:
Thanks for the review @eloquence! I'll address the comments tonight or tomorrow.

gonzalo-bulnes:
Applied the suggestions and rebased. Waiting for potential `href` additions :slightly_smiling_face:

docs/development/l10n.rst
rmol: Haven't had coffee yet, but I'm thinking if a URL were to vary like that, it would probably be done via a `{url}` placeholder, so the translators wouldn't have to worry about it. I'd keep `href` in the list of things not to touch.

eloquence:
(Rebased.)

eloquence (approved):
Thanks @gonzalo-bulnes, per @rmol's comment re: `href` translation I am merging as-is; we can add further clarification if needed.

PR #125, opened 2021-01-07
Update permanently redirected links

### LINKED ISSUE #124: Update the links that are permanently redirected to reduce linkcheck noise
## Describe the change

The output of `make docs-linkcheck` currently contains 60+ warnings that correspond to **permanent redirects**. Those are rendered in red, probably to signify that since the redirection is _permanent_ they could/should be updated.

This doesn't impact documentation users, because the redirection will take them to the page they're looking for.

However, it impacts the maintainers of the documentation by attracting their attention to links that are not broken, and ultimately making the linkcheck output longer and less readable.

## User Research Evidence

None. However, it has happened to me to unintentionally ignore a warning because I mistook it for one that I had intentionally ignored (too) many times before. And that seems to be a common effect that unnecessary warnings have on people.

Since linkcheck provides the updated URL as part of the warnings, I'd say this is a low-hanging fruit in terms of effort.

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

Updates the links that are flagged as **redirected permanently** by `make docs-linkcheck`, in order to reduce the amount of warnings in the output of the tool.

* Fixes #124 

## Testing

- [ ] Review that _each_ link points to the same place as before. :grimacing: There are a lot of them, but I can't think of a better way to review that looking at them one by one.
- [ ] Review the decisions made for the few broken links (e.g. the link to the NUC5i5MYHE kit that doesn't seem to be documented by Intel anymore)

## Release 

- [ ] Run `make linkcheck`
- [ ] Preview the docs locally and verify that the links are rendered as expected

## Checklist (Optional)

### DIFF
--- a/docs/development/contributing.rst
+++ b/docs/development/contributing.rst
@@ -67,7 +67,7 @@ project and its components, including the graphical `SecureDrop Client <https://
 
 When you're ready to share your work with the SecureDrop team for review, submit
 a `pull request
-<https://help.github.com/categories/collaborating-with-issues-and-pull-requests/>`__
+<https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests>`__
 with the proposed changes. :doc:`Tests <testing_securedrop>` will run
 automatically on GitHub.
 
@@ -169,7 +169,7 @@ Forum Moderators and Support
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Those running a production instance of SecureDrop are encouraged to `read the
-support documentation <https://securedrop-support.readthedocs.io/>`__ to get
+support documentation <https://support-docs.securedrop.org/>`__ to get
 help from the `Freedom of the Press Foundation <https://freedom.press>`__. For
 less sensitive topics such as running a demo or getting help to understand a
 concept, a `public forum section <https://forum.securedrop.org/c/support>`__ is
--- a/docs/hardware.rst
+++ b/docs/hardware.rst
@@ -686,7 +686,7 @@ Network Switch
 This is optional, for people who are using a firewall with less than 4
 ports.  Any old switch with more than 3 ports will do, such as the
 `5-port Netgear ProSafe Ethernet Switch
-<http://www.amazon.com/NETGEAR-ProSafe-Gigabit-Ethernet-Desktop/dp/B0000BVYT3/>`__.
+<https://www.amazon.com/NETGEAR-Ethernet-Unmanaged-Protection-GS105NA/dp/B0000BVYT3>`__.
 
 .. _printers_tested_by_fpf:
 
--- a/docs/hardware.rst
+++ b/docs/hardware.rst
@@ -529,7 +529,7 @@ Intel 5th-gen NUC
 ~~~~~~~~~~~~~~~~~
 
 We previously recommended the
-`NUC5i5MYHE <https://www.intel.com/content/www/us/en/products/boards-kits/nuc/kits/nuc5i5myhe.html>`__,
+`NUC5i5MYHE <https://web.archive.org/web/20190102175756/https://www.intel.com/content/www/us/en/products/boards-kits/nuc/kits/nuc5i5myhe.html>`__,
 however, it has now reached end-of-life. We will continue to support and
 test SecureDrop on this hardware, but if you are building a new SecureDrop
 instance we recommend using 7th- or 8th-generation NUCs instead.
--- a/docs/hardware.rst
+++ b/docs/hardware.rst
@@ -171,7 +171,7 @@ SecureDrop system. Each admin and each journalist needs a two-factor
 device. We currently support two options for two-factor authentication:
 
 * Your existing smartphone with an app that computes TOTP codes
-  (e.g. FreeOTP `for Android <https://play.google.com/store/apps/details?id=org.fedorahosted.freeotp>`__ and `for iOS <https://itunes.apple.com/us/app/freeotp-authenticator/id872559395>`__).
+  (e.g. FreeOTP `for Android <https://play.google.com/store/apps/details?id=org.fedorahosted.freeotp>`__ and `for iOS <https://apps.apple.com/us/app/freeotp-authenticator/id872559395>`__).
 
 * A dedicated hardware dongle that computes HOTP codes (e.g. a
   `YubiKey <https://www.yubico.com/products/yubikey-hardware/yubikey/>`__).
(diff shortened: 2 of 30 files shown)


gonzalo-bulnes:
For what is worth, here is the [script I used to perform the bulk of the update](https://gist.github.com/gonzalo-bulnes/e18cf01e64153871244245dd4ebcfd47). It could be re-used.

The script doesn't spare reviewing the links one by one and is not perfect, but it makes the search and replace quicker.

docs/development/contributing.rst
```diff
@@ -80,7 +80,7 @@ Technical Writers
 ~~~~~~~~~~~~~~~~~
 
 Technical writers and editors are invited to review the `documentation
-<https://docs.securedrop.org/>`__ and fix any mistakes in accordance with the
+<https://docs.securedrop.org/en/stable/>`__ and fix any mistakes in accordance with the
```
eloquence, line 83: In this case, and others where we currently link directly to a root of an RTD site like https://docs.securedrop.org/ and https://support-docs.securedrop.org/, I would prefer to keep doing so. That way, if we want to change the default version for any reason, we do not have to  update those URLs.

docs/rebuild_admin.rst
```diff
@@ -6,7 +6,7 @@ Rebuilding an *Admin Workstation* USB
           rebuild your *Admin Workstation*, please contact FPF through the
           `SecureDrop Support Portal`_.
 
-.. _SecureDrop Support Portal: https://securedrop-support.readthedocs.io/en/latest/
+.. _SecureDrop Support Portal: https://support-docs.securedrop.org/en/latest/en/latest/
```
eloquence, line 9: This is an invalid URL that 404s (notice repetition of `/en/latest`)

docs/update_tails_usbs.rst
```diff
@@ -94,4 +94,4 @@ If you run into issues, you can always restore your data from the backup device
 
 If you continue to have problems, you can contact us through the `SecureDrop Support Portal`_.
 
-.. _SecureDrop Support Portal: https://securedrop-support.readthedocs.io/en/latest/
+.. _SecureDrop Support Portal: https://support-docs.securedrop.org/en/latest/en/latest/
```
eloquence, line 97: This is an invalid URL that 404s (notice repetition of `/en/latest`)

docs/hardware.rst
```diff
@@ -528,11 +528,11 @@ Previous Server Recommendations
 Intel 5th-gen NUC
 ~~~~~~~~~~~~~~~~~
 
-We previously recommended the
-`NUC5i5MYHE <https://www.intel.com/content/www/us/en/products/boards-kits/nuc/kits/nuc5i5myhe.html>`__,
-however, it has now reached end-of-life. We will continue to support and
+We previously recommended the NUC5i5MYHE, however,
+it has now reached end-of-life. We will continue to support and
```
eloquence, line 532: It might be nice to include a Wayback link here, with the link label "archived copy of specifications" or similar:
https://web.archive.org/web/20190102175756/https://www.intel.com/content/www/us/en/products/boards-kits/nuc/kits/nuc5i5myhe.html

eloquence (changes requested):
Thanks @gonzalo-bulnes, that's very helpful and should reduce URL fragility. One comment regarding the docs URLs in particular, a note re: the NUC5 link, and a couple of broken links; otherwise LGTM. Will need one more rebase before merge.

docs/development/contributing.rst
gonzalo-bulnes: This makes sense 👍 

I'll take a look at what settings are available in _linkcheck_ in case there is a way to avoid rendering the permanent redirect warnings in these cases. (I saw some ignored links there may be finer-grained options.)

docs/rebuild_admin.rst
gonzalo-bulnes: Good catch, thank you! I thought I had caught all of those, I missed them when proof reading : /

docs/hardware.rst
gonzalo-bulnes: I didn't think of that! Good idea!

gonzalo-bulnes:
I'll address the comments and rebase. Thank you for the review @eloquence !

gonzalo-bulnes:
I fixed the issues highlighted in the comments and rebased to update the branch.

docs/network_firewall.rst
```diff
@@ -28,14 +28,14 @@ If you are new to pfSense or firewall management in general, we
...
-   Wiki <https://doc.pfsense.org/index.php/Main_Page>`__
+   Wiki <https://docs.netgate.com/pfsense/en/latest/index.html>`__
 -  `pfSense: The Definitive
    Guide <https://www.amazon.com/pfSense-Definitive-Christopher-M-Buechler/dp/0979034280>`__
 
    -  *Note:* This guide is now slightly out of date, although we found
       it to be a useful reference in the past. To get the latest version of
       this book, you need to become a `pfSense Gold
-      Member <https://www.pfsense.org/our-services/gold-membership.html>`__.
+      Member <https://www.pfsense.org/our-services/#gold-membership>`__.
```
gonzalo-bulnes, line 38: Note that Gold Membership doesn't seem to be offered by pfSense anymore, and `make linkcheck` reports that the `#gold-membership` anchor is not found.

I found references to a "version 2.1" published in 2013, but I didn't find any in the pfSense docs and I don't know how it can be obtained anymore.

eloquence: I'm inclined to remove this link altogether; it was flagged as "slightly out of date" more than 5 years ago (https://github.com/freedomofpress/securedrop/commit/8be932c1932d22fc75e684573c67329271e1a84f). @zenmonkeykstop or others; any thoughts on other pfSense resources we may want to cross-reference here?

zenmonkeykstop: https://docs.netgate.com/pfsense/en/latest/index.html is a good resource, which is linked elsewhere apparently - I'd agree with just removing this one along with the obsolete info about memberships.

gonzalo-bulnes:
Except for the misspelled link (https://github.com/freedomofpress/securedrop-docs/pull/127),  `make docs-linkcheck` is now green.

docs/network_firewall.rst
```diff
@@ -28,14 +28,7 @@ If you are new to pfSense or firewall management in general, we
 recommend the following resources:
```
eloquence, line 28: resources->resource

singular now

eloquence: (fixed in 3cf2ffeb29b033792b07f1cd5ddaa125b28b448a)

eloquence:
LGTM! Pushed one tiny fix-up commit. As with the other PR, please rebase to pick up the LFS objects, and then we can land it.

docs/network_firewall.rst
gonzalo-bulnes: Thanks!

eloquence (approved):
LGTM, thanks much for the hard work on this!

Issue #122, opened 2021-01-05
Update v2/v3 Onion Page Text


ninavizz:
## Describe the change
Hi Gang! Until I learn Git a little more, I can’t really do PRs on my own, but wanted to flag the request for some edits to [this page](https://docs.securedrop.org/en/stable/v3_services.html?highlight=v3) in the Docs. [Edit suggestion, in comment below](https://github.com/freedomofpress/securedrop-docs/issues/122#issuecomment-755089890). These edits are important for the user experience and comprehension of info to be smooth, when folx click on the link in the new JI banners telling them to upgrade.

## How will this impact users?
The page is confusing and a challenging read at the moment, for users arriving on the page from the banner we'll be pushing live in `1.7.0`. It's important to communicate with users what we need them to know, first; and everything, later.

The "Order Of Consequence" with the text on this page, seems to be:
1. v2 is going to be depreciated, and you have until April 30th to not be screwed by this—know that.
2. You need to update your onion address on your Landing Page if you've already done the migration but haven't done that part.
3. If you don't do the above, your Sources will not know where to go.
4. Here is why all of this is happening.
5. Here is how to do it all, if you haven't already.

The topical 101 in the first paragraph of a page, is also quite confusing. Per #123 I'm recommending creating a new item in the glossary, and moving the existing content there.

With the increased urgency of getting users to Update, it feels extra important to prioritize user engagement for this page when users link to it from the `1.7.0` JUI banner.

## User Research Evidence
Nope! Someday, though. :)
Recommendations from UX SME follow general best-practices for writing instructional content that users need to follow.

ninavizz:
## Suggested edits, below.
1. Includes linked use of word "Onion Services"
2. Removes confusing use of word "currently" to disambiguate v2 from v3 services. New/Old mental models should be used with this, not "Current/Former," as they're both still live and therefore technically current atm. 
3. Re-order text so page makes more sense and promotes correct urgency, when read from banner on `1.7.0`
4. Updates text above examples of v2 and v3 service address examples. Existing text is structured to require users "to think" instead of following more common UI patterns for qualifying information. Including words like "Cryptography Primitives" is discouraged, as so few users will understand what that means for the specificity to have value.


> **IMPORTANT:** Tor project is deprecating support for v2 [onion services](#). SecureDrop will remove support for v2 onion services as part of its 2.0.0 release, planned for February 2021. If you are currently using v2 onion services, they will become unreachable at that point.
> 
[...]

PR #128, opened 2021-01-08
Add Onion Service entry to Glossary

### LINKED ISSUE #123: Create "Onion Services" item in Glossary
## Describe the change
Create a new item in the glossary for Onion Services. Cited as need for #122 

Text from existing Onion Services blab in docs, reads as follows—and seems like a perfect starting point. I would also suggest adding-in something about the `.onion` suffix and unreadable URLs that are 16 (v2) or 56 (v3) characters (that they exist, not details about v2 vs v3 or depreciation). Frame as "New vs Previous" not "Current vs Old," as both remain live today.

> Tor onion services provide anonymous inbound connections to websites and other servers exclusively over the Tor network. For example, SecureDrop uses onion services for the Journalist and Source Interface websites, as well as for administrative access to the servers in SSH-over-Tor mode.

Also

> The longer v3 addresses provide stronger cryptographic primitives (algorithms) than v2 onion services, and include redesigned protocols that guard against service information leaks on the Tor network.

Will make blabs on technical Onion stuff more readable and easy to consume, if the 101 stuff can live in a glossary listing vs in page content. Is also useful for all users to be able to reference from user guides.

## User Research Evidence
Nope. That would be so nice, though. :)

UX SME notes the above suggestions follow general best practice for content consumption, and advises y'all to please do. UX best practices informed by research by practitioners. Nope, we don't do RTD or otherwise have shared best practice knowledge documented as copiously well as developers document coding languages and rules. 

Generalized resources for best practices wrt content writing, IA, etc., for docs, in UX repo.

## Additional context
Just wanting to make Docs more actionable and content easier to consume. :)

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

Add an "Onion Service" entry to the Glossary.

Includes a "Onion Service v2" (sub-)entry that outlines the differences between v3 and v2 onion services and recommends the use of the most recent version. The intention is that this sub-entry can be removed when support for v2 onion services is dropped, but that removal is not strictly necessary either (to avoid adding strong imperatives to future releases).

Fixes #123

## Testing

- [ ] Please proof read (I'm not an native English speaker!)
- [ ] Check that the definition makes sense for someone who has no other context
- [ ] Check that the definition plays well with [advice given elsewhere regarding onion services](https://github.com/freedomofpress/securedrop-docs/issues/122)
- [ ] Check that the sub-section looks all right in the docs table of contents and the permalink is usable (`.../glossary.html#onion-service-v2`)

## Release 

* I can't think of any special consideration for releasing this change into the stable version of the documentation.

## Checklist (Optional)

- [ ] ~~Doc link linting (`make docs-linkcheck`) passed~~ (no new links)

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -19,8 +19,9 @@ Application Server
 ------------------
 The *Application Server* runs the SecureDrop application. This server hosts both
 the website that sources access (the *Source Interface*) and the website that
-journalists access (the *Journalist Interface*). Sources, journalists, and
-admins may only connect to this server using Tor.
+journalists access (the *Journalist Interface*). Both are published
+through an *onion service* because sources, journalists, and admins
+may only connect to this server using Tor.
 
 
 Export Device
@@ -92,6 +93,37 @@ Monitor Server
 The *Monitor Server* keeps track of the *Application Server* and sends out an
 email alert if something seems wrong. Only system admins connect
 to this server, and they may only do so using Tor.
+
+
+Onion Service
+-------------
+
+Tor onion services provide anonymous inbound connections to websites and other
+servers exclusively over the Tor network. For example, SecureDrop uses onion
+services for the *Journalist Interface* and *Source Interface* websites,
+as well as for administrative access to the servers in SSH-over-Tor mode.
+
+Onion services can be accessed by clicking a link or pasting the onion service
+address into Tor Browser. For example,
+``sdolvtfhatvsysc6l34d65ymdwxcujausv7k5jk4cy5ttzhjoi6fzvyd.onion`` is the onion
+service address for the SecureDrop website.
+
+Read more about `onion services in Tor's glossary
+<https://support.torproject.org/onionservices/>`__.
+
+Onion Service v2
+""""""""""""""""
+
+The third generation of onion services (v3) provides stronger cryptographic
+algorithms than v2 onion services, and includes redesigned protocols that
+guard against service information leaks on the Tor network.
+
+Support for v2 onion services will start to be removed from SecureDrop in
+February 2021, and we strongly recommend :doc:`upgrading SecureDrop
+instances to v3 onion services <v3_services>` as soon as possible.
+
+Distinguishing between different generations of onion services is easy:
+v3 addresses are longer (56 characters) than v2 addresses (16 characters).
 
 
 OSSEC Alert Public Key


gonzalo-bulnes:
**Notes**

- While the original issue called for the creation of a "Onion Services" entry, I opted for the singular "Onion Service" by consistency with "Export Device", "Transfer Device", "Submission Key".

- I first thought of a more detailed recommendation, then decided against because the future tense would become inaccurate when SecureDrop 2.0.0 is released (this section should be removed as part of the release, but it may be best not to make it an imperative?):

> ~SecureDrop support for v2 onion services will end when SecureDrop 2.0.0 is released.~ It is strongly recommended that only v3 onion services are used with SecureDrop 1.7.0 and forward.

gonzalo-bulnes:
> I would also suggest adding-in something about the .onion suffix and unreadable URLs that are 16 (v2) or 56 (v3) characters [...]

@ninavizz Do the SecureDrop onion address example and the v3 vs v2 distinction tip address this in a satisfactory way?

> Frame as "New vs Previous" not "Current vs Old," as both remain live today.

I tried to avoid "New vs Previous" altogether to make the copy more future-proof. Do you think saying "New vs Previous" would make things clearer though? (The docs will need to be maintained anyway, so there are limits to the usefulness of future-proofing in the "clarity vs maintainability" trade-off. And I think we can agree that clarity should prevail in this case.)


> UX SME notes the above suggestions follow general best practice for content consumption, and advises y'all to please do. 

My intent was to structure the glossary entry as follows:

1. What is an onion service? (that's your copy actually :P thank you for the precise guidelines!)
2. What does it look like? How do I use it? (It's not that different from any other link, provided you use Tor Browser)
3. Sub-section (You may hear of "v2"...)
4. What is that about? (Different versions.)
5. Why do I care? (SecureDrop support, security)
6. How do I know I'm doing the right thing? (URL length, thinking that any glossary reader could bring their admins' attention to this.)
7. I tried to place an example of use of "onion service" in context.

I'd add a caveat on language as I'm not an English native speaker.

What's your UX SME view on the result? What would you change?

docs/glossary.rst
```diff
@@ -94,6 +95,23 @@ email alert if something seems wrong. Only system admins connect
...
+-------------
+
+Tor onion services provide anonymous inbound connections to websites and other servers exclusively over the Tor network. For example, SecureDrop uses onion services for the *Journalist Interface* and *Source Interface* websites, as well as for adminsitrative access to the servers in SSH-over-Tor mode.
+
+Onion services can be accessed by clicking a link or pasting the onion service address into Tor Browser. For example, `sdolvtfhatvsysc6l34d65ymdwxcujausv7k5jk4cy5ttzhjoi6fzvyd.onion <https://sdolvtfhatvsysc6l34d65ymdwxcujausv7k5jk4cy5ttzhjoi6fzvyd.onion>`__ is SecureDrop's onion service address.
+
+Onion Service v2
+""""""""""""""""
+
+The third generation of onion services (v3) provides stronger cryptographic primitives (algorithms) than v2 onion services, and include redesigned protocols that guard against service information leaks on the Tor network.
```
gonzalo-bulnes, line 108:
```suggestion
The third generation of onion services (v3) provides stronger cryptographic primitives (algorithms) than v2 onion services, and includes redesigned protocols that guard against service information leaks on the Tor network.
```

💡 @ninavizz This can be a convenient way to amend copy. In the "Files changed" tab of this PR, near the beginning of any line, you can make a little "+" button appear to leave a comment. In the comment composer, there is an icon representing a file with a +/- symbol that allows to **insert a suggestion**. Clicking it copies the line as-is and allows you to edit it at will. Once you post the comment, a diff is automatically generated to highlight your changes 🙂 

It only works on line that have already been changed (that appear in the "Files changed" tab), but the result can be committed directly using the GitHub interface, so it's quite handy for larger re-wordings.

gonzalo-bulnes: **Extra level**: if you drag the little "+" icon when leaving a comment, you can leave a comment on multiple lines at once. ✨

docs/glossary.rst
```diff
@@ -94,6 +95,23 @@ email alert if something seems wrong. Only system admins connect
...
+
+It is strongly recommended that only v3 onion services are used with SecureDrop 1.7.0 and forward.
```
gonzalo-bulnes, line 110:
```suggestion
It is strongly recommended that only v3 onion services are used from SecureDrop 1.7.0 onwards.
```

eloquence:
(Rebased.)

docs/glossary.rst
```diff
@@ -19,8 +19,9 @@ Application Server
 ------------------
 The *Application Server* runs the SecureDrop application. This server hosts both
 the website that sources access (the *Source Interface*) and the website that
-journalists access (the *Journalist Interface*). Sources, journalists, and
-admins may only connect to this server using Tor.
+journalists access (the *Journalist Interface*). Both are published
+through an *onion service* because Sources, journalists, and admins
```
eloquence, line 23: Sources->sources (mid-sentence)

docs/glossary.rst
```diff
@@ -94,6 +95,23 @@ email alert if something seems wrong. Only system admins connect
 to this server, and they may only do so using Tor.
 
 
+Onion Service
+-------------
+
+Tor onion services provide anonymous inbound connections to websites and other servers exclusively over the Tor network. For example, SecureDrop uses onion services for the *Journalist Interface* and *Source Interface* websites, as well as for adminsitrative access to the servers in SSH-over-Tor mode.
```
eloquence, line 101: adminsitrative-> administrative

docs/glossary.rst
```diff
@@ -94,6 +95,23 @@ email alert if something seems wrong. Only system admins connect
...
+
+Onion services can be accessed by clicking a link or pasting the onion service address into Tor Browser. For example, `sdolvtfhatvsysc6l34d65ymdwxcujausv7k5jk4cy5ttzhjoi6fzvyd.onion <https://sdolvtfhatvsysc6l34d65ymdwxcujausv7k5jk4cy5ttzhjoi6fzvyd.onion>`__ is SecureDrop's onion service address.
```
eloquence, line 103: We generally prefer not to link to onion services, because such links are clickable in a clearnet browser, which may cause DNS leakage before the request fails. While this is just an example address, we should follow the same convention here. Note that this example address is not reachable using `https://` (HTTPS is not required for onion services, which provide their own layer of encryption, though it can be useful as a way to assert ownership).

eloquence: I would also change "SecureDrop's onion service address" to "the onion service address for the SecureDrop website", just to be minimally ambiguous (which SecureDrop?).

docs/glossary.rst
eloquence: We can strengthen this recommendation now:

> Support for v2 onion services will start to be removed from SecureDrop in February 2021, and we strongly recommend [upgrading SecureDrop instances to v3 onion services](https://docs.securedrop.org/en/stable/v3_services.html) as soon as possible.

docs/glossary.rst
eloquence: (FYI, Nina is taking a break until later in January and may not be able respond here.)

IMO if we unpack "cryptographic primitives (algorithms)" we may as well just say "cryptographic algorithms".

docs/glossary.rst
eloquence: It may be useful to cross-link Tor's own glossary entry here, which may also offer some inspiration for making the language more accessible:
https://support.torproject.org/onionservices/

eloquence (changes requested):
Thanks @gonzalo-bulnes - a few comments/suggestions inline. (Note that I rebased onto latest `main`, so I suggest updating your branch before working on it.)

docs/glossary.rst
gonzalo-bulnes: :bulb: Oh, quite a few things here, thank you for the explanations @eloquence.

docs/glossary.rst
gonzalo-bulnes: @eloquence I noticed that Tor's glossary doesn't apply the same [caution when linking onion addresses](https://github.com/freedomofpress/securedrop-docs/pull/128#discussion_r555407001).

> For example, the DuckDuckGo onion is https://... (link)

Because of that, I added the cross-reference in a separate commit that you can keep, squash or remove when performing the rebase.

gonzalo-bulnes:
Thank you for the suggestions & explanations @eloquence! I applied them and left you a couple of comments.

By the way, I keep amending my commits and breaking lines at ~80 characters. Please let me know if you prefer a series of `--fixup` commits. There isn't too much text in this case but I know it can be easier to review distinct changes? (I could make them ready to `--autosquash` if that makes handling easier.)

docs/glossary.rst
gonzalo-bulnes: > Including words like "Cryptography Primitives" is discouraged, as so few users will understand what that means for the specificity to have value. I moved that bit to suggested text for the Onion Services glossary item. (@ninavizz's notes in #122)

I made the change to "cryptographic algorithms", though on second thought, I'm wondering if "**security features**" wouldn't be more accessible, even in the glossary.

eloquence:
> By the way, I keep amending my commits and breaking lines at ~80 characters. Please let me know if you prefer a series of --fixup commits.

Either is fine with me, thanks. :) I have a soft preference for "separate commits first, then squash later", but experience shows that the "squash later" step is often forgotten.

eloquence:
LGTM! I added one small cleanup commit. If all looks good to you as well, please rebase once more, and we can get this in.

(I would rebase and merge, but the `main` branch includes new LFS objects which cannot be pushed to forks without permission; if you add me as a collaborator on your fork, I can rebase in future cases like this.)

eloquence (approved):
LGTM, thank you. :)

Issue #36, opened 2020-10-20
Add docs on v3 onion service restore-from-backup story


rocodes:
## Description

As we are encouraging the adoption of v3 onion services, we should improve on the documentation on restoring from a backup:

- with v3 only
- with v2+v3 (based on a current support portal request) 

## User Research Evidence

Open support ticket

## User Stories

As an administrator, I want clear instructions on how to restore from a backup if my instance supports v3 onion services.

eloquence:
(Removed from board as #36, which is expected to fully resolve, is tracked there.)

PR #133, opened 2021-01-13
Improve Backup/Restore/Migration docs, add v3 instructions 

rocodes:
Include backup and restore instructions for v3 onion services. Add hardware migration instructions for v3 onion services and for cases where onion credentials may not be restored.

## Status

Work in progress

## Description of Changes

* Description:  Update Restore/Migration docs to include steps for v3 services. Slight edits to condense existing Backup/Restore copy. WIP: would appreciate review. 

* Fixes Issue #36

## Testing
manual review 

## Release 
none

## Checklist (Optional)

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -695,6 +695,8 @@ When troubleshooting, ensure you are on the latest version of SecureDrop
 in your *Admin Workstation*. This is done by accepting the update
 when prompted at boot in the GUI that appears.
 
+.. _troubleshooting_admin_connectivity: 
+
 I can't SSH into my servers over Tor from my Admin Workstation. What do I do?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -739,6 +741,7 @@ into your Admin Workstation, you should first perform the following troubleshoot
    Ensure this key is available using ``ssh-add -L``. If you see the output
    "This agent has no identities." then you need to add the key via ``ssh-add``
    prior to SSHing into the servers.
+
 
 I got a unusual error when running ``./securedrop-admin install``. What do I do?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/backup_and_restore.rst
+++ b/docs/backup_and_restore.rst
@@ -1,28 +1,25 @@
 Back Up, Restore, Migrate
 =========================
 
-
-There are a number of reasons why you might want to backup and restore a
-SecureDrop installation. Maintaining periodic backups is generally a good
-practice to guard against data loss. In the event of hardware failure on
-the SecureDrop servers, having a recent backup will enable you to redeploy
-the system without changing Onion URLs, recreating journalist accounts,
+Maintaining regular backups helps guard against data 
+loss and hardware failure. Having a recent backup will allow you to redeploy
+SecureDrop without changing onion URLs, recreating journalist accounts,
 or losing previous submissions from sources.
 
 .. note:: Only the *Application Server* is backed up and restored, including
           historical submissions and both *Source Interface* and *Journalist
-          Interface* URLs. The *Monitor Server* needs to be configured from
-          scratch in the event of a hardware migration.
+          Interface* URLs. The *Monitor Server* needs to be 
+          configured from scratch in the event of a hardware migration.
 
 Minimizing Disk Use
 -------------------
 
 Since the backup and restore operations both involve transferring *all* of
 your SecureDrop's stored submissions over Tor, the process can take a long time.
-To save time and improve reliability for the transfers, take a moment to clean up
-older submissions in the *Journalist Interface*. As a general practice, you should
-encourage Journalists to delete regularly unneeded submissions from the *Journalist
-Interface*.
+
+Encouraing journalists to regularly delete older, unneeded submissions from 
+the *Journalist Interface* will save time and improve reliability when 
+doing backups.
 
 .. tip:: Although it varies, the average throughput of an onion service is
          about 300 kB/s, or roughly 2 hours for 2GB. Plan your backup and
@@ -31,6 +28,15 @@ Interface*.
 You can use the following command to determine the volume of submissions
 currently on the *Application Server*: log in over SSH and run
 ``sudo du -sh /var/lib/securedrop/store``.
+
+Compare the output of this command (which approximates the size of 
+a backup archive) to the amount of free space on your Tails persistent volume
+via Tails' **Disks** utility to ensure you have sufficient space to perform
+a backup.
+
+If you find you cannot perform a backup or restore due to this constraint,
+and have already deleted old submissions from the *Journalist Interface*,
+contact us through the `SecureDrop Support Portal`_.
 
 .. note:: Submissions are deleted asynchronously and one at a time, so
           if you delete a lot of submissions through the *Journalist
@@ -44,53 +50,29 @@ currently on the *Application Server*: log in over SSH and run
             sudo journalctl -u securedrop_rqworker
 
 
-If you find you cannot perform a backup or restore due to this constraint,
-and have already deleted old submissions from the *Journalist Interface*,
-contact us through the `SecureDrop Support Portal`_.
-
 .. _SecureDrop Support Portal: https://support-docs.securedrop.org/
+
+.. _backing_up:
 
 Backing Up
 ----------
 
-Open a **Terminal** on the *Admin Workstation* and ``cd`` to your clone
-of the SecureDrop git repository (usually ``~/Persistent/securedrop``).
-Ensure you have a tagged SecureDrop release checked out, version 0.4 or
-later. (You can run ``git describe --exact-match`` to verify that you have the
-right source checked out.)
+Open a **Terminal** on the *Admin Workstation* and ``cd`` to 
+``~/Persistent/securedrop``.
 
-.. note:: The backups are stored in the *Admin Workstation* persistent volume.
-          **Verify that you have enough space to store the backups
-          before running the backup command.**
-
-          You can use the ``du`` command described earlier to get the
-          approximate size of the backup file (since the majority of the backup
-          archive is the stored submissions), and Tails' **Disks** utility to
-          see how much free space you have on your persistent volume.
+Run ``git describe --exact-match`` to verify that you have a tagged SecureDrop release, 0.4 or newer, checked out. 
 
 Check Connectivity
 ''''''''''''''''''
 
-First, verify that your *Admin Workstation* is able to run Ansible and connect to
+Verify that your *Admin Workstation* is able to run Ansible and connect to
 the SecureDrop servers.
 
 .. code:: sh
 
    ssh app uptime
 
-If this command fails (usually with an error like "SSH Error: data could not be
-sent to the remote host. Make sure this host can be reached over ssh"), you need
-to debug your connectivity before proceeding further. Make sure:
-
-* Ansible is installed
-* the *Admin Workstation* is connected to the Internet
-* Tor starts successfully
-* The appropriate onion service configuration files are present in
-  ``~/Persistent/securedrop/install_files/ansible-base`` and the
-  ``./securedrop-admin tailsconfig`` command completes successfully
-
-If Ansible is not installed, or ``./securedrop-admin tailsconfig`` fails, see
-:doc:`configure_admin_workstation_post_install` for detailed setup instructions.
+If this command fails, see :ref:`Troubleshooting <troubleshooting_admin_connectivity>`.
 
 Create the Backup
 '''''''''''''''''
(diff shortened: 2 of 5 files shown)


docs/backup_and_restore.rst
```diff
@@ -123,67 +105,216 @@ Restoring
...
+time of the backup.
 
-* ``app-source-ths``
-* ``app-journalist-aths``
-* ``app-ssh-aths``
+.. _migrating:
 
-Once SSH access to the servers has been established (or if using SSH over
-local network), Onion URLs for the *Source Interface* and *Journalist Interfaces*
-can be fetched using the installer:
```
rocodes, line 172: @zenmonkeykstop: I did not use `./securedrop-admin install` to fetch onion URLs _to_ the Admin Workstation. I copied aths/auth-private files into place and ran tailsonfig, so I have provisionally removed this step. If I'm missing something, please let me know.

eloquence:
(Taking a first pass.)

docs/admin.rst
```diff
@@ -572,6 +574,11 @@ into your Admin Workstation, you should first perform the following troubleshoot
    "This agent has no identities." then you need to add the key via ``ssh-add``
    prior to SSHing into the servers.
 
+#. **Ensure that Ansible is installed.** If Ansible is not installed, 
+   or ``./securedrop-admin tailsconfig`` fails, see
+   :doc:`configure_admin_workstation_post_install` for detailed setup instructions.
```
eloquence, line 579: What motivates this addition? We already suggest running `./securedrop-admin tailsconfig` above; the command will instruct the admin to run `./securedrop-admin setup` if the virtual environment has not been initialized. The instructions you link to mainly ask the admin to run `./securedrop-admin tailsconfig`, so this feels somewhat circular.

docs/backup_and_restore.rst
```diff
@@ -1,12 +1,9 @@
...
 
-
-There are a number of reasons why you might want to backup and restore a
-SecureDrop installation. Maintaining periodic backups is generally a good
-practice to guard against data loss. In the event of hardware failure on
-the SecureDrop servers, having a recent backup will enable you to redeploy
-the system without changing Onion URLs, recreating journalist accounts,
+Maintaining regular backups helps guard against data 
+loss and hardware failure. Having a recent backup will allow you to redeploy
+SecureDrop without changing Onion URLs, recreating journalist accounts,
```
eloquence, line 6: Yay for removing fluff sentences :)

docs/backup_and_restore.rst
```diff
@@ -123,67 +105,216 @@ Restoring
...
+   your file and appending ``.onion``, as shown below:
+
+   .. code:: sh
+
+      ssh-keygen -f "/home/amnesia/.ssh/known_hosts" -R "app_ssh_url.onion"
+
+   Then, edit the file called ``tor_v3_keys.json``, located in 
+   ``install_files/ansible-base`` to replace the two ``app_journalist`` and 
+   two ``app_ssh`` values in the file with their respective values 
+   that you preserved in the ``tor_v2_keys.json`` file from the previous installation.
```
eloquence, line 267: `tor_v2_keys.json`->`tor_v3_keys.json`

docs/backup_and_restore.rst
```diff
@@ -123,67 +105,216 @@ Restoring
...
+or you are performing a hardware migration, see :ref:`Migrating <migrating>`. 
+
+If your instance has switched from v2 to v3 onion services since your
+most recent backup, see the section 
+:ref:`on mismatched Tor configurations <restore-with-mismatched-tor-config>`, 
+or  `contact Support`_.
 
 .. note:: The backup strategy used for SecureDrop versions prior to 0.3.7
           created encrypted archives with the extension ``.zip.gpg``.
           You can safely remove those files once you've created the ``.tar.gz``
```
eloquence, line 128: ^ SecureDrop 0.3.7 is a long time ago, I'd nominate this block for removal while we're overhauling this chapter.

docs/backup_and_restore.rst
```diff
@@ -123,67 +105,216 @@ Restoring
...
+
+      cat install_files/ansible-base/app-ssh-aths
+
+   This file is in the format ``HidServAuth app_ssh_url Secret # client: admin``.
+   Type the following command, replacing ``app_ssh_url`` with the .onion URL 
+   found in the file above:
+
+   .. code:: sh
+
+      ssh-keygen -f "/home/amnesia/.ssh/known_hosts" -R "app_ssh_url"
```
eloquence, line 245: Because this command is essentially identical for both v2 and v3, I would suggest consolidating this a bit further: 1) Obtain the SSH service .onion address for the Application Server (instructions vary between v2 and v3); 2) Remove the old onion service form the list of known hosts (instructions are identical once you have the .onion address).

Regarding the sample command, it's not obvious to me at a glance that `app_ssh_url` is meant to be replaced by the user. Further up we use `<your_backup_file>` to indicate that the filename is to be substituted by the user, which has the problem of using reserved shell symbols `<` and `>` but is a pretty common pragma regardless. I'm open to other ideas, but I do think we want it to stand out a bit more.

docs/backup_and_restore.rst
```diff
@@ -123,67 +105,216 @@ Restoring
...
-Then rerun ``./securedrop-admin tailsconfig`` to update the *Admin Workstation*
-to use the restored Onion URLs again. See :doc:`configure_admin_workstation_post_install`
-for detailed instructions.
+#. :ref:`Back up the existing installation <backing_up>`.
 
-Migrating
----------
+#. **Preserve important credentials**: Before re-installing SecureDrop, you 
+   will need to preserve some important credential files from the old 
+   installation on the *Admin Workstation*. They should be temporarily saved
```
eloquence, line 170: I would change "They should be .." to active voice, e.g., "Save these credentials in .."

docs/backup_and_restore.rst
```diff
@@ -123,67 +105,216 @@ Restoring
...
+
+   **Instances using v3 onion services:**
+
+   - ``app-sourcev3-ths``
+   - ``app-ssh.auth_private``
+   - ``app-journalist.auth_private``
+   -  ``tor_v3_keys.json``
+
+
+   Once you have preserved those files, you may either delete or rename the
```
eloquence, line 194: I'm not sure we even want to enumerate deletion as an option at this stage, in case any contents of the directory are later needed for diagnostics. I would suggest just offering the ``securedrop.old`` step here.

docs/backup_and_restore.rst
```diff
@@ -123,67 +105,216 @@ Restoring
...
+   Once you have preserved those files, you may either delete or rename the 
+   SecureDrop project root directory to something like ``securedrop.old`` and
+   delete it when the migration is complete. 
+
+#. **Prepare the new servers.** Re-clone the SecureDrop repository, and 
+   :doc:`reinstall SecureDrop from scratch <install>`. 
+
+   During configuration (``./securedrop-admin sdconfig``), set up the
+   new SecureDrop installation with the same onion 
+   services version(s) as your previous server configuration. Proceed through
```
eloquence, line 203: I think it may be desirable to spell this out a bit more to avoid confusion about what "onion service version(s)" means in this context. How about something like this:

> During the configuration step (`./securedrop-admin sdconfig`), you will be prompted to choose whether to enable v2 onion services, v3 onion services, or both. Make sure that your selection is the same as it was in the previous server configuration.

docs/backup_and_restore.rst
```diff
@@ -123,67 +105,216 @@ Restoring
...
+   **v3 onion services**
+
+   * ``app-sourcev3-ths``
+   * ``app-journalist.auth_private``
+   * ``app-ssh.auth_private``
+
+#. **Reconfigure Tor connections**: 
+   Run ``./securedrop-admin tailsconfig`` to update the *Admin Workstation*
+   to use the restored Onion URLs again. See :doc:`configure_admin_workstation_post_install`
+   for further instructions.
```
eloquence, line 294: Perhaps add a recommendation to verify connectivity here?

docs/backup_and_restore.rst
```diff
@@ -123,67 +105,216 @@ Restoring
...
-If you are using the same *Admin Workstation* to do a restore from a previous
-backup, it should already be there because it was placed there by the backup
-command. Otherwise, you should copy the backup archive that you wish to restore to
-``install_files/ansible-base``.
+If you are restoring data onto an existing instance whose onion URLs have not
+changed (for example, for data recovery purposes), see :ref:`Restoring Data on an
+Existing SecureDrop Installation <restore_data>`.
+
+If you are restoring data onto an instance whose onion URLs *have* changed,
+or you are performing a hardware migration, see :ref:`Migrating <migrating>`.
```
eloquence, line 119: The phrase "onion URLs have changed" seems ambiguous to me (relative to what?), and I think it could be difficult for an admin to choose between the "Migrating" pointer and the following "mismatched Tor config" pointer based on how this is currently worded. As I understand it, we want to cover two scenarios in this chapter:

- restoring data for an existing, working installation
- reinstalling and then restoring from a backup, either for hardware/platform migration or for disaster recovery

Depending on when the last backup was taken, both scenarios may require the use of `--preserve-tor-config` to avoid clobbering a working Tor config on the server. If that's accurate, I would suggest structuring the chapter into these two sections at a high level, and explaining the `--preserve-tor-config` behavior in context where appropriate.

eloquence:
Thanks @rocodes for filling in some very important detail for these instructions! Left some comments based on my initial review; the structural question may be easier to talk through synchronously esp. with @zenmonkeykstop to figure out what we want to address in the docs revision now, vs. defer to later.

docs/admin.rst
rocodes: Makes sense. I did pull it from the other troubleshooting steps and put it here as a last resort diagnostic (as in, if none of the above steps work, double-check that ansible is even installed). I wasn't sure how to put that. But if it reads circular we can remove it.

rocodes:
@eloquence these comments are great, thank you. It was a little challenging figuring out how to clarify the steps here. First round of revisions coming later today.

docs/admin.rst
rocodes: Removed this instruction in 151175d

docs/backup_and_restore.rst
rocodes: yikes, fixed in 151175d

docs/backup_and_restore.rst
rocodes: Removed in 151175d

docs/backup_and_restore.rst
rocodes: Updated in 151175d - let me know if this wording works for you.

docs/backup_and_restore.rst
rocodes: Added in 151175d

docs/backup_and_restore.rst
rocodes: Thank you for this feedback! In 151175d I've described the instructions as you indicate, and I've moved the section on `--preserve Tor config` to a section of "Additional Instructions" at the end. (Eventually I will include a small section in that Additionals section on rebuilding the monitor server, as per e.g. https://github.com/freedomofpress/securedrop/issues/5311, but for now the only thing in there is the `preserve Tor config` piece.)

docs/backup_and_restore.rst
```diff
@@ -123,67 +105,213 @@ Restoring
...
+.. note:: These steps require physical access to the servers. 
+         
+Moving a SecureDrop installation to new hardware consists of:
+
+  - Backing up the existing installation and preserving its Tor credentials from the *Admin Workstation* 
+  - Installing the same version of SecureDrop on the new hardware;
+  - Restoring the backup to the new installation;
+  - Repairing Tor credentials for the new installation. 
+
+Detailed Migration Instructions
```
zenmonkeykstop, line 155: These instructions are a bit complex and branching - it might make sense to split them into 3 sections as follows:

1) v3-only - this should be simple: 
- copy `tor_v3_keys.json` to the fresh  clone of securedrop, 
- then proceed with a fresh install 
- and then restore from the backup

2) v2+v3 - actually, this will be the same as v3-only, as who cares about the v2 ssh setup, it's not used. You can copy app's `*ths` files as below if you want the `ssh app-legacy` and `ssh mon-legacy` commands to work
3) v2-only - this one is probably the most complex - you will need to 
- copy app's `*ths` files from old install, do a fresh install, 
- restore from backup (which will fail with an error that says to run tailsconfig)
- copy app's `*ths` files into place, 
- run tailsconfig, 
- then run restore again. 

The trick here is that you don't have to rebuild mon's config for v2, you can just set up a new ssh config. Also for the configs with a v3 element, all you should need is the `tor_v3_keys.json` file.

(Also I may be missing steps here, it's been a while!)

zenmonkeykstop: Ooh also you'll need to copy keys from `securedrop.old` and you might as well copy `site-specific` while you're in there.

zenmonkeykstop (changes requested):
I *think* the migration instructions could be simpler. Either way I'd recommend restructuring them to treat the v2, v3, v2+3 scenarios separately.

docs/backup_and_restore.rst
zenmonkeykstop: Or you could skip the new clone and just move files you don't want out of the way....

eloquence: So, what's the scenario in which we want to support them preserving v2 as part of a migration/reinstall? Should we just make it explicit that they should migrate to v3 while they're there? These docs didn't use to exist at all, if we add them now, it feels like we might as well make them v3-only.

zenmonkeykstop: Hypothetically? Someone may have a downed v2 instance with an old backup available, and they may want to restore it before migrating. Or they have some other reason unique to them for going right up to the wire on v2. But adding a "Data-only migration to v3" option might not be a bad idea. something like:
---
If you want to restore the data from a backup on a fresh instance with new onion service addresses:

- complete the fresh install first
- copy the backup file FILENAME to install_files/ansible base
- run `./securedrop-admin restore --preserve-tor-config FILENAME`
---
would do the trick. (this is v2/v3 only but if we disable new v2 installs next release that's a strong hint.)

docs/backup_and_restore.rst
```diff
@@ -1,28 +1,25 @@
...
-older submissions in the *Journalist Interface*. As a general practice, you should
-encourage Journalists to delete regularly unneeded submissions from the *Journalist
-Interface*.
+
+Encouraing journalists to regularly delete older, unneeded submissions from 
+the *Journalist Interface* will save time and improve reliability when 
+doing backups.
 
 .. tip:: Although it varies, the average throughput of an onion service is
          about 300 kB/s, or roughly 2 hours for 2GB. Plan your backup and
```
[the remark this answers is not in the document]
rocodes: https://metrics.torproject.org/onionperf-throughput.html?start=2020-08-01&end=2021-01-21&server=onion I'm guestimating here, but I'll update it to ~ 90 mins for 2 GB, based on ~3 Mbps

docs/backup_and_restore.rst
rocodes: So the "data-only migration to v3" section is in here, more opaquely referred to as  "Restoring Data Without Restoring Tor Configuration", and is for the case of instances with a v2 backup file who want to recover data (ahem, if certain deadlines are missed). 

I think there's an argument for skipping the v2-only hardware migration instructions, and telling users to contact Support. As soon as we stop allowing v2 onion services on fresh installs, the v2 migration instructions won't work. However, since they are written up I'm including them for now anyone says otherwise, and we can remove them in a few weeks' time.

eloquence: @rocodes: 
> However, since they are written up I'm including them for now 

Sounds reasonable; perhaps we can just add a blurb at the top of the section, e.g.:

> If you are migrating or restoring an installation using v2 onion services, please be aware that support for them will be removed from SecureDrop starting in March 2021. We encourage you to migrate to v3 onion services as part of any migration. To do so, reinstall SecureDrop with v3 onion services enabled and restore your data without overwriting the generated Tor configuration **[link to section]**. The instructions for users of v2 onion services below will be removed in a future release.

rocodes:
I've addressed @zenmonkeykstop's minor edits (for typos etc) in d913113, and also taken a stab at the restructuring/splitting into sections via onion service. His workflow is slightly different than the one I used (and I think his is better), but I haven't tested each path yet so I still consider this WIP.

docs/backup_and_restore.rst
rocodes: Added in note on depreciation 254b0d9, and also split out migration instructions by onion services version.

docs/backup_and_restore.rst
```diff
@@ -1,28 +1,25 @@
...
 -------------------
 
 Since the backup and restore operations both involve transferring *all* of
 your SecureDrop's stored submissions over Tor, the process can take a long time.
-To save time and improve reliability for the transfers, take a moment to clean up
-older submissions in the *Journalist Interface*. As a general practice, you should
-encourage Journalists to delete regularly unneeded submissions from the *Journalist
-Interface*.
+
+Encouraing journalists to regularly delete older, unneeded submissions from
```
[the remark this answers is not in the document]
rocodes: oop. fixed & thank you.

docs/backup_and_restore.rst
rocodes: Resolved because we now instruct users to copy in `site-specific`

eloquence:
Thanks @rocodes. I've taken another visual spin through, and it looks great to me. 

@rocodes and @zenmonkeykstop: What level of additional technical review would you recommend for these changes? My main concern is that stepping through a full migration, let alone all 3, will take significant time. Perhaps we should decouple this specific docs update from the 1.7.0 release timetable?

rocodes:
I would like to step through a full migration on Monday before asking for anyone else's review, and I think it's doable. I think it's reasonable to decouple this from 1.7.0 and land it even a day or two after release day.

rocodes:
@zenmonkeykstop: on the v3 scenario, I'm not having success with the workflow you suggested. Specifically, after the `restore`, I still need to reboot the app server and repair ssh creds, subbing in the original `app-ssh.auth_private`, similar to the v2 workflow. `tor_v3_keys.json` alone doesn't seem to do the trick for me. Let's connect tomorrow to make sure I'm not missing anything.

rocodes:
Here are the v3 steps that worked for me @zenmonkeykstop, could use your eye before I write it up.

* copy tor_v3_keys.json, pubkey files, and site-specific to the fresh clone of securedrop, and preserve `app-ssh.auth_private` file elsewhere (for now)
* proceed with a fresh install, run tailsconfig, then run `ssh app` and `ssh mon` 
* copy in the backup file, restore from the backup, which will work but then time out
* reboot the app server (or restart tor via console) 
* copy in `app-ssh.auth_private` file, rerun `tailsconfig`
* run `install`, then run `tailsconfig` again.

docs/backup_and_restore.rst
```diff
@@ -123,67 +107,308 @@ Restoring
...
-* ``app-ssh-aths``
+   Copy in the following files from ``~/Persistent/securedrop/install_files/ansible-base``:
 
-Once SSH access to the servers has been established (or if using SSH over
-local network), Onion URLs for the *Source Interface* and *Journalist Interfaces*
-can be fetched using the installer:
+   -  ``tor_v3_keys.json``
+   - All ``.asc`` files (these correspond to your *Submission Public Key*, your OSSEC alerts public key, and (if configured) your journalist alerts public key
+   - The file ``~/Persistent/securedrop/install_files/ansible-base/group-vars/all/site-specific``
+   - (optional) ``app-ssh-aths`` and ``mon-ssh-aths``, if you wish to reinstate the ``app-ssh-legacy`` and ``mon-ssh-legacy`` commands
```
zenmonkeykstop, line 235: mon isn't affected by the restore, so the monitor server after the restore will still have its  address from the fresh install. This means there's some divergence between how to handle v2 and v3 in this case. You can just retain the fresh install mon-ssh-aths in the v2 case, but you would need to update the original tor_v3_keys.json, adding in the mon-ssh keypair from the fresh install's tor_v3_keys.json, before running either tailsconfig (which will update the auth_private file) or install (which will update the client auth files on the server).

docs/backup_and_restore.rst
```diff
@@ -123,67 +107,308 @@ Restoring
...
+   Run ``./securedrop-admin tailsconfig`` to update the *Admin Workstation*
+   to use the restored Onion URLs again, and verify connectivity to the 
+   servers and the *Source* and *Journalist Interfaces*. 
+   See :doc:`configure_admin_workstation_post_install`
+   for further instructions.
+
+#. **Delete temporary files**: Remove any temporary files or directories
+   you created as you performed this migration, including the 
+   ``securedrop.old`` directory.
+
```
zenmonkeykstop, line 383: There's one other restore/migrate case to consider based on changes in https://github.com/freedomofpress/securedrop/pull/5677 - namely the v2+v3 -> v3 one (previously didn't work but now unblocked)

zenmonkeykstop (changes requested):

docs/backup_and_restore.rst
```diff
@@ -44,53 +50,31 @@ currently on the *Application Server*: log in over SSH and run
...
 
-.. note:: The backups are stored in the *Admin Workstation* persistent volume.
-          **Verify that you have enough space to store the backups
-          before running the backup command.**
-
-          You can use the ``du`` command described earlier to get the
-          approximate size of the backup file (since the majority of the backup
-          archive is the stored submissions), and Tails' **Disks** utility to
-          see how much free space you have on your persistent volume.
+Run ``git describe --exact-match`` to verify that you have a tagged SecureDrop
```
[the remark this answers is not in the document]
eloquence: I would suggest conservatism until after April 30, given that out-of-date workstations will not have this change when it's needed, and so the check should still be performed manually during migrations. Agree that the 0.12.0 reference should go though.

docs/backup_and_restore.rst
```diff
@@ -123,67 +107,308 @@ Restoring
...
+Moving a SecureDrop installation to new hardware consists of:
+
+  - Backing up the existing installation and preserving credentials from the *Admin Workstation*; 
+  - Installing the same version of SecureDrop on the new hardware and copying credentials into place;
+  - Restoring the backup to the new installation.
+
+The instructions differ depending on which onion services version 
+your instance is using.
+
+* :ref:`Instances using v3 onion services <migrate_v3>`
```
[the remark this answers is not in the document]
rocodes: fixed in 5c29677, and addressed all other comments in same commit

docs/backup_and_restore.rst
```diff
@@ -1,37 +1,43 @@
...
+doing backups.
 
 .. tip:: Although it varies, the average throughput of an onion service is
-         about 300 kB/s, or roughly 2 hours for 2GB. Plan your backup and
+         about 3 Mbps, or roughly 90 minutes for 2GB. Plan your backup and
          restore accordingly.
 
 You can use the following command to determine the volume of submissions
 currently on the *Application Server*: log in over SSH and run
 ``sudo du -sh /var/lib/securedrop/store``.
```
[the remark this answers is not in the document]
rocodes: fixed in  5c29677 (& all such code-blocks also)

rocodes:
@zenmonkeykstop: I think I've addressed all your comments.  I also stepped through the v3 instructions again last night. 

- For v3 services, there is no mucking around with `tor_v3_keys.json` _if_ you copy it into place before running the clean install, as detailed in these instructions. 
- I've separated the common instructions into an `includes`. Honestly, the only thing that differs about the v2 and v3 instructions is the "hey, if you want, you can also save the old `app-ssh-aths` file so your legacy commands work." That will be irrelevant real soon anyways, so we will be able to cut this document in half pretty soon.
- To address the v2v3 xenial -> v3 focal scenario, I've made a note on the v2/v3 instructions that, when restoring onto Focal, the correct path is just to follow the v3 steps. Does that work for you?  I've also written up Focal migration instructions (separate PR, tk later today) that indicate which migration documents to follow (basically, the "v3 docs" or the "preserve-tor-config" docs).

docs/backup_and_restore.rst
rocodes: See my comment- I have addressed this in a note in front of the v2+v3 instructions, indicating that migrating to Ubuntu 20.04 --> follow the v3-only instructions.

zenmonkeykstop:
^^ one thing to note about patched-in changes below - the migration instructions include a tag verification/checkout step, which currently refers to 1.7.1. It should be 1.8.0,  this is updated by update_version.sh which will be run prior to pushing to stable.

rocodes:
Closing in favour of #168

PR #137, opened 2021-01-21
add note about how we update the builder image

creviera:
## Status

Ready for review

## Description of Changes

Add a note in the docs about how we ignore the test failure that checks if the container that builds our debian packages is up-to-date with the latest packages. The test failure is expected and future improvements to this process can be made after our Focal upgrade, see  see https://github.com/freedomofpress/securedrop/issues/4533.

### DIFF
--- a/docs/development/dockerbuildmaint.rst
+++ b/docs/development/dockerbuildmaint.rst
@@ -1,12 +1,21 @@
-Docker Build Maintenance
-========================
+.. _build_container:
 
-Get your Quay account squared away
------------------------------------
-The container that performs builds of Debian packages is version controlled in
-a docker repository at **quay.io/freedomofpress/sd-docker-builder**.
-There are tight restrictions over who can make edits here. If you have permissions
-to do so, you'll need to make sure your local docker client has credentials to push.
+Build container
+===============
+We use a Docker build container to build our debian packages for SecureDrop (via ``make build-debs``
+in the ``securedrop`` Github repository root directory). We keep images of this our container in a
+Docker repository at **quay.io/freedomofpress**. The images are organized by Ubuntu release
+version. For instance, you can find the images for Xenial at
+**quay.io/freedomofpress/sd-docker-builder-xenial** and, for Focal, at
+**quay.io/freedomofpress/sd-docker-builder-focal**.
+
+Maintaining images of our bulid container for each release is our way of recording the exact version
+of each dependency used to build our production debian packages for SecureDrop.
+
+Who can update the build container?
+===================================
+There are tight restrictions over who can make edits to our Docker repository. If you have
+permissions to do so, you'll need to make sure your local Docker client has credentials to push.
 
 * First login into your quay.io account via the web-portal at https://quay.io/
 * Drill into your **Account settings** via the upper right drop-down (where your username is)
@@ -15,12 +24,16 @@ to do so, you'll need to make sure your local docker client has credentials to p
   obtained from the previous step.
 * Proceed with update instructions
 
+Updating the build container
+============================
+We know the build container needs to be updated when **test_ensure_no_updates_avail** fails during
+``make build-debs`` in the ``securedrop`` Github reprository root directory. This test fails if any
+of the dependencies required to build the debian packages have security updates. If you have access
+rights to push to quay.io, then you can build and push a new container to the Quay repository by
+following the steps below.
 
-Performing container updates
-----------------------------
-If one of the dependencies requires security updates, the build may fail at
-**test_ensure_no_updates_avail** . If you have access rights to push to quay.io,
-here is the process to build and push a new container:
+.. note:: The reason we don't update the container at runtime is that we use the container image as
+          a way of recording our build environment.
 
 .. code:: sh
 
--- a/docs/development/dockerbuildmaint.rst
+++ b/docs/development/dockerbuildmaint.rst
@@ -9,7 +9,7 @@ version. For instance, you can find the images for Xenial at
 **quay.io/freedomofpress/sd-docker-builder-xenial** and, for Focal, at
 **quay.io/freedomofpress/sd-docker-builder-focal**.
 
-Maintaining images of our bulid container for each release is our way of recording the exact version
+Maintaining images of our build container for each release is our way of recording the exact version
 of each dependency used to build our production debian packages for SecureDrop.
 
 Who can update the build container?
--- a/docs/development/release_management.rst
+++ b/docs/development/release_management.rst
@@ -43,11 +43,12 @@ Pre-Release
    goal is to make sure we test against the lastest Tails release, including release candidates,
    so that we can report bugs early to Tails.
 
-#. Check if there is a new stable release of Tor that can be QAed and released as part of the
-   SecureDrop release. Also check for any new release candidates so that we're aware of any
-   upcoming major bug fixes and communicate them to the team. You can find releases by checking the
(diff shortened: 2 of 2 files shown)


docs/development/release_management.rst
```diff
@@ -127,23 +128,32 @@ Pre-Release
...
-   #. Build the packages with ``make build-debs``.
-   #. Build logs should be saved and published according to the `build
+   b. Build the packages with ``make build-debs``.
+
+     .. note:: Since we don't update the container that builds the debian packages at runtime, you
+               might end up seeing ``test_ensure_no_updates_avail`` fail in the build output. This
+               tells us that our builder image needs to be updated to ensure we are using the latest
+               security packages. You can ignore this test failure for release candidates, but
+               create a ticket in the ``securedrop`` repository so that we can update the builder
+               image before building the packages for the final release.
```
eloquence, line 138: It it necessary to create an issue to update the builder image (historically it doesn't look like we've done so), or is it standard operating procedure to always update it before building the packages for the final release?

creviera: I think you're right. Perhaps creating an issue to update the builder image is overkill. We could just add in the "Release Process" section below a step for updating the builder image if this test fails. So the note here could say something more like:
```
     .. note:: Since we don't update the container that builds the debian packages at runtime, you
               might end up seeing ``test_ensure_no_updates_avail`` fail in the build output. This
               tells us that our builder image needs to be updated to ensure we are using the latest
               security packages. You can ignore this test failure for release candidates, because
               it will be updated as part of the final release (linking to the release process section).
```

eloquence: That sounds right to me -- @emkll, does that make sense to you as well?

emkll: Yes, @creviera that language looks good to me. We could also link to the build instructions, since the release management documentation is generally developer-facing: docs/development/dockerbuildmaint.rst


Note that in the "Final Release" section of Release issues, we have a check box in place to ensure the builder is up-to-date, so i agree that opening an issue may be redundant here:
```
 Ensure builder in release branch is updated and/or update builder image
```

creviera: okay i also update docs/development/dockerbuildmaint.rst so this is ready again for review

docs/development/dockerbuildmaint.rst
```diff
@@ -1,12 +1,21 @@
...
+Build container
+===============
+We use a Docker build container to build our debian packages for SecureDrop (via ``make build-debs``
+in the ``securedrop`` Github repository root directory). We keep images of this our container in a
+Docker repository at **quay.io/freedomofpress**. The images are organized by Ubuntu release
+version. For instance, you can find the images for Xenial at
+**quay.io/freedomofpress/sd-docker-builder-xenial** and, for Focal, at
+**quay.io/freedomofpress/sd-docker-builder-focal**.
+
+Maintaining images of our bulid container for each release is our way of recording the exact version
```
eloquence, line 12: typo: bulid->build

docs/development/dockerbuildmaint.rst
```diff
@@ -15,12 +24,16 @@ to do so, you'll need to make sure your local docker client has credentials to p
...
+rights to push to quay.io, then you can build and push a new container to the Quay repository by
+following the steps below.
 
-Performing container updates
-----------------------------
-If one of the dependencies requires security updates, the build may fail at
-**test_ensure_no_updates_avail** . If you have access rights to push to quay.io,
-here is the process to build and push a new container:
+.. note:: The reason we don't update the container at runtime is that we use the container image as
+          a way of recording our build environment.
```
eloquence, line 36: Could equivalent security properties be achieved by including the exact version information in the build logs automatically, as suggested in https://github.com/freedomofpress/securedrop/issues/4533#issuecomment-705717195 ? If so, it does not seem a compelling argument to include in favor of the current process.

eloquence:
Looks good to me (one smol typo), but since I'm not directly familiar with the builder image update process, I would appreciate review from a SecureDrop maintainer before merging.

docs/development/dockerbuildmaint.rst
```diff
@@ -1,12 +1,21 @@
...
-There are tight restrictions over who can make edits here. If you have permissions
-to do so, you'll need to make sure your local docker client has credentials to push.
+Build container
+===============
+We use a Docker build container to build our debian packages for SecureDrop (via ``make build-debs``
+in the ``securedrop`` Github repository root directory). We keep images of this our container in a
+Docker repository at **quay.io/freedomofpress**. The images are organized by Ubuntu release
+version. For instance, you can find the images for Xenial at
+**quay.io/freedomofpress/sd-docker-builder-xenial** and, for Focal, at
+**quay.io/freedomofpress/sd-docker-builder-focal**.
```
conorsch, line 10: We could make these bolded URLs actual URLs to be a bit easier on folks, but that's not a change that came in via this PR.

docs/development/dockerbuildmaint.rst
conorsch: The docs presented here are definitely accurate in terms of the process we actually adhere to. I remain interested in getting the SD core debs fully reproducible, so we can stop fussing with the build container so much. After Focal, that seems a reasonable goal again, consistent with discussion in https://github.com/freedomofpress/securedrop/issues/4533. 

All that to say, the "note" as written here LGTM!

docs/development/release_management.rst
```diff
@@ -43,11 +43,12 @@ Pre-Release
...
-   SecureDrop release. Also check for any new release candidates so that we're aware of any
-   upcoming major bug fixes and communicate them to the team. You can find releases by checking the
-   `Tor blog <https://blog.torproject.org/category/tags/stable-release>`_. If there is a new
-   stable release, file an issue and upgrade Tor following these steps:
+#. Check the Tor blog for new
+   `release candidates <https://blog.torproject.org/category/tags/release-candidate>`_ and new
+   `stable releases <https://blog.torproject.org/category/tags/stable-release>`_. Let the team
+   know about any new release candidates during the SecureDrop release process in case there are
+   critical bug fixes. For a new stable release, file an issue and upgrade Tor following these
+   steps:
```
conorsch, line 51: Much clearer, thank you for including the explicit URLs!

docs/development/release_management.rst
```diff
@@ -127,23 +128,30 @@ Pre-Release
...
       within 15 minutes.
 
-   .. warning:: Only commit deb packages with an incremented version number: do not clobber existing
-                packages. That is, if there is already a deb called e.g.
-                ``ossec-agent-3.6.0-amd64.deb`` in ``main``, do not commit a new version of this
-                deb.
+     .. warning:: Only commit deb packages with an incremented version number: do not clobber
+                  existing packages. That is, if there is already a deb called e.g.
+                  ``ossec-agent-3.6.0-amd64.deb`` in ``main``, do not commit a new version of this
+                  deb.
```
conorsch, line 150: Locally, the way I determine this is by running `git status` and checking if a file was _modified_ by my changes. If so, I run `git reset <file>` on each of them, and commit only the new packages. That's something CI could check for us, even in the dev-lfs repo, that might be a little friendlier to developers. Weighed against continued progress toward reproducibility, when we can trust a bot to file the PR, I very much prefer these docs you've got here.

conorsch (approved):
Looks great! Thanks for preparing, @creviera. There are a few procedures in here we'll surely revisit in the near future, but no question that the changes you've made here are more accurate, and should be included in the dev docs right away.

Issue #151, opened 2021-02-10
Add screenshots for disabling secure boot on nuc8


creviera:
## Describe the change

This is a reminder to take screenshots of the bios for disabling secure boot on nuc8.

eloquence:
The need to disable SecureBoot on NUCs is currently entirely undocumented; @creviera has committed to adding screenshots and brief instructions. We also discussed that it would be worthwhile to track & discuss SecureBoot support for SecureDrop core (cf. related issue for SecureDrop Workstation: https://github.com/freedomofpress/securedrop-workstation/issues/602)

PR #158, opened 2021-02-23
add disable secure boot instructions

creviera:
## Description

Add instructions to disable Secure Boot (until we make this so that it's no longer a requirement).

Can't remember if I had to do this for nuc7, but it might make sense to move the instructions to the top as a check for all nucs.

* Fixes https://github.com/freedomofpress/securedrop-docs/issues/151

### DIFF
--- a/docs/hardware.rst
+++ b/docs/hardware.rst
@@ -413,16 +413,19 @@ to disable wireless functionality:
 
 |NUC8 VisualBios1|
 
-While in the BIOS, you should also navigate to **Advanced > Security** in the
-and disable SGX support, which not used by SecureDrop and may be targeted by
-active CPU exploits:
+- navigate to **Advanced > Security** in the BIOS and disable SGX support, which is not used by 
+  SecureDrop and may be targeted by active CPU exploits:
 
 |NUC8 VisualBios2|
+
+- navigate to **Advanced > Boot > Secure Boot** and uncheck the **Secure Boot** checkbox:
+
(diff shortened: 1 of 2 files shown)


zenmonkeykstop:
I think it's good as is - the pending PR for focal installs includes a general section on BIOS settings (with Secure Boot info), and the NUC7s will probably age out of the reccos list soon.

docs/hardware.rst
```diff
@@ -413,16 +413,19 @@ to disable wireless functionality:
 
 |NUC8 VisualBios1|
 
-While in the BIOS, you should also navigate to **Advanced > Security** in the
-and disable SGX support, which not used by SecureDrop and may be targeted by
-active CPU exploits:
+- navigate to **Advanced > Security** in the BIOS and disable SGX support, which is not used by 
+  SecureDrop and may be targeted by active CPU exploits:
```
eloquence, line 417: nice catch on the `is not` :)

eloquence:
`git checkout` on this branch with LFS installed notes that the binary is not tracked via LFS:

```
Encountered 1 file(s) that should have been pointers, but weren't:
	docs/images/hardware/nuc8_visualbios_secureboot.png
```

If LFS is installed, I believe `git add` and amending the existing commit should resolve.

eloquence (approved):

eloquence:
LGTM, thanks for resolving LFS issue :)

PR #154, opened 2021-02-17
Update server installation instructions to use Focal instead of Xenial

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

- Updates server OS install instructions to use Ubuntu 20.04 LTS
- Removes workarounds for NUC8 Ethernet support, as they're not required with Focal's base kernel.

## Testing
- Are the docs clear and correct?
- Following them, do you get a valid SD install?

## Checklist (Optional)

- [x] Doc linting (`make docs-lint`) passed locally
- [ ] Doc link linting (`make docs-linkcheck`) passed **not tested**
- [x] You have previewed (`make docs`) docs at http://localhost:8000

Issue #135, opened 2021-01-19
Add Ubuntu 20.04 install and migration instructions


eloquence:
As part of the Ubuntu 20.04 transition (https://github.com/freedomofpress/securedrop/issues/4768) we need to update our installation guide to be accurate for Ubuntu 20.04. We also need to provide migration instructions for administrators. For the latter, we may be able to build on Ro's generalized migration docs in #133.

eloquence:
Last time around we also had a "xenial prep" doc (https://github.com/freedomofpress/securedrop/pull/4046); will create similar specifically mentioning v3 as a preparatory step folks can complete, towards this issue.

eloquence:
@zenmonkeykstop has offered to sketch out the migration process as part of the current sprint, informed by the generalized migration procedures @rocodes wrote up in #133. This can be done in a wiki page for now; our initial goal is to ensure that we have a migration procedure we can test for the following configurations:

- v2-only (v2 onion service config will not be restored once https://github.com/freedomofpress/securedrop/pull/5677 is merged)
- v3+v2 or v3-only

eloquence:
Completed mainly via #169, #168 and #154.

PR #160, opened 2021-02-26
Focal migration (take 2)

rocodes:
## Status

Work in progress - feedback welcome

## Description of Changes

* Description: Adds migration instructions, mostly by referring to the soon-to-be-merged hardware migration guidelines. Removes the "take a backup" and the v3 onion services guidelines from the preparatory steps, and indicates 2 possible migration paths. Renames the page from 'focal prep' to 'focal migration'. 
* Towards #135 (in conjunction with #154) 

## Testing
Visual review 

## Release 
Wait for #133 to land, will require references to that PR 
Wait for #154, will refer to Ubuntu 20.04 instructions as well  

## Checklist (Optional)


rocodes:
Curses, this is a clean branch so I'm not sure why I'm seeing those extra merge commits. Okay, well 
3234e41 is what I was trying to put up here, just for some preliminary feedback on whether we like the general structure/approach before I spend too much time on it.  Ignore the git cruft and let me know how you feel about the structure.
The migration guide relies heavily on the 2 PRs I referenced, and honestly they make up the bulk of the migration instructions.

docs/upgrade/focal_migration.rst
```diff
@@ -0,0 +1,240 @@
...
+
+At a high level, the migration process consists of:
+
+- Taking a backup of the current instance
+- Installing Ubuntu 20.04 LTS and SecureDrop from scratch
+- Applying the backup
+
+Instances that already have :doc:`v3 onion services <../v3_services>` enabled 
+and follow our migration guide will be able to 
+preserve their existing *Source* and *Journalist Interface* onion URLs.
```
eloquence, line 29: As a reader, that left me with the question "what if I'm not running v3?". I know this is clarified later in the doc, but maybe add another sentence here, e.g.:
> If you are still running v2 onion services, you must switch to v3 as part of the migration at the latest, and your onion addresses will change.

docs/upgrade/focal_migration.rst
```diff
@@ -0,0 +1,240 @@
...
+this command in a terminal:
+
+.. code:: sh
+
+ ssh app apt-cache policy securedrop-app-code
+
+SecureDrop servers are updated automatically with the latest release version.
+If your servers are running an old version, this indicates a major configuration
+problem, and you may need to reinstall SecureDrop. In that case, please
+:ref:`contact us <contact_us>` for assistance.
```
eloquence, line 99: I realize this is language from the existing prep guide, but on re-reading it here, it seems to me we should probably drop the "you may need to reinstall SecureDrop" here. They'll need to reinstall SecureDrop regardless of their current version in order to migrate; if their current server is so badly misconfigured that it's not even on 1.7.1, that may just imply that a careful review is required to determine if the data can be safely moved over to the new install. So I would suggest just dropping that clause, and keeping the "please contact us".

docs/upgrade/focal_migration.rst
```diff
@@ -0,0 +1,240 @@
...
+Migration
+~~~~~~~~~
+
+.. _migration_standard:
+
+Standard Migration Procedure
+----------------------------
+(For SecureDrop instances already using v3 onion services)
+
+#. :doc:`Take a backup of the current instance <../backup_and_restore>`.
```
eloquence, line 190: If there are steps that are identical for both migration scenarios (e.g., make a backup, download/verify Ubuntu 20.04 media), it may make sense to split those into their own section for preparatory steps. I do recognize that means that earlier links won't be able to go directly to the path-specific section.

eloquence:
Thanks @rocodes, what's there so far is looking very good to me on visual review. Note that this PR currently includes some merge commits and an unrelated commit; just ping me if you'd like help resolving that.

docs/upgrade/focal_migration.rst
rocodes: For downloading/verifying Ubuntu 20.04, yes, I can add that to the preparatory steps. The reason I think that making a backup belongs during the actual migration window is it's time-sensitive; ideally the instance would be backed up and then taken offline immediately, so that no submissions fall through the cracks.

eloquence: Yes, that makes sense; I meant that the backup paragraph is duplicated in the two sections for the different paths, and if we can consolidate migration steps that are the same regardless of v2/v3 status, that might make it easier to maintain. But understand if you prefer to err on the side of just having each migration procedure be fully self-contained.

rocodes:
Replaced by #169

PR #168, opened 2021-03-08
Add migration instructions for v3 instances and clarify migration paths

rocodes:
## Status

Ready for review 
Replaces #133 

## Description of Changes

* Description:  Add backup and restore information for v3 instances. Add detailed hardware migration information. Include information on migration with `preserve-tor-config` flag. 

## Testing
* visual review

## Release 
* n/a 

## Checklist (Optional)


docs/includes/v3_backup_steps_common.rst
```diff
@@ -0,0 +1,146 @@
+.. note::  The instructions below assume that you are using the same *Admin Workstation*
+   that was used to manage your old instance. If you are using a new *Admin
+   Workstation* you will need to copy the directory ``~amnesia/Persistent/securedrop``
+   from the old workstation to the new workstation (using a *Transfer Device*)
+   before proceeding.
+
+#. If you have not already done so, :ref:`back up the existing installation <backing_up>`. The instructions below assume that the backup has been created and renamed ``sd-backup-old.tar.gz``.
```
eloquence, line 7: I would just add this as a step, "Rename the backup file to .."

docs/includes/v3_backup_steps_common.rst
```diff
@@ -0,0 +1,146 @@
...
+
+      cd ~/Persistent/securedrop
+      export SD_OLD="~/Persistent/sd.bak/install_files/ansible-base"
+      export SD_NEW="~/Persistent/securedrop/install_files/ansible-base"
+      cp $SD_OLD/group_vars/all/site-specific $SD_NEW/group_vars/all/site-specific
+      cp $SD_OLD/tor_v3_keys.json $SD_NEW/tor_v3_keys.json
+      cp $SD_OLD/sd-backup-old.tar.gz $SD_NEW/sd-backup-old.tar.gz
+
+   You will also need to copy the old instance's *Submission Public Key*,
+   *Ossec Alert Public Key*, and, if configured, the *Journalist Alert Public Key*.
```
eloquence, line 79: Ossec->OSSEC

docs/includes/v3_backup_steps_common.rst
rocodes: This one I think I'm going to leave (this naming is part of @zenmonkeykstop's patch), because elsewhere it refers to assuming filenames like `ossec.asc`, `journalist.asc`, `ca.crt`, etc, so I think it's clear that those are placeholder names.

docs/includes/v3_backup_steps_common.rst
```diff
@@ -0,0 +1,146 @@
...
+   Assuming that these are named ``SecureDrop.asc``, ``ossec.asc``, and
+   ``journalist.asc`` respectively, run the following commands:
+
+   .. code:: sh
+
+      cp $SD_OLD/SecureDrop.asc $SD_NEW/SecureDrop.asc
+      cp $SD_OLD/ossec.asc $SD_NEW/ansible-base/ossec.asc
```
eloquence, line 86: Since `$SD_NEW` includes `ansible-base` I believe it is duplicated here.

docs/includes/v3_backup_steps_common.rst
```diff
@@ -0,0 +1,146 @@
...
+      cp $SD_OLD/journalist.asc $SD_NEW/ansible-base/journalist.asc
```
eloquence, line 87: As above.

docs/includes/v3_backup_steps_common.rst
eloquence: You could use the shorter `$SD_NEW/` for all commands that use the same filename as the original.

docs/includes/v3_backup_steps_common.rst
```diff
@@ -0,0 +1,146 @@
...
+   unreachable.
+
+#. Copy the old instance's v3 onion service details into place on the
+   *Admin Workstation* and repair SSH access using the Terminal commands:
+
+   .. code:: sh
+
+      cd ~/Persistent/securedrop
+      cp $SD_OLD/app-{journalist,ssh}.auth_private $SD_NEW/
+      cp $SD_OLD/app-sourcev3-ths $SD_NEW
```
eloquence, line 142: Trailing slash missing here (both will work just fine; the trailing slash is nice to ensure it errors out in case the directory is not present as expected).

docs/backup_and_restore.rst
```diff
@@ -101,89 +88,257 @@ When you are ready to begin the backup, run
...
+      cd ~/Persistent
+      git clone https://github.com/freedomofpress/securedrop
+
+#. Verify that the current release tag was signed with the release signing key:
+
+   .. code:: sh
+
+      cd ~/Persistent/securedrop/
+      git fetch --tags
+      git tag -v 1.7.1
```
eloquence, line 234: This version string is not encapsulated by your changes to `update_version.sh` (it will remain at 1.7.1 after bumping the version).

docs/backup_and_restore.rst
```diff
@@ -101,89 +88,257 @@ When you are ready to begin the backup, run
...
+      and you should **not** proceed with the installation. If this happens,
+      please contact us at securedrop@freedom.press.
+
+
+   Verify that each character of the fingerprint matches what is on the
+   screen of your workstation. If it does, you can check out the new release:
+
+   .. code:: sh
+
+      git checkout 1.7.1
```
eloquence, line 255: As above.

docs/backup_and_restore.rst
```diff
@@ -101,89 +88,257 @@ When you are ready to begin the backup, run
...
+
+   .. important::
+      If you see the warning ``refname '1.7.1' is ambiguous`` in the
```
eloquence, line 258: As above.

docs/backup_and_restore.rst
rocodes: I noticed this (the changes you are remarking on are from @zenmonkeykstop 's patch). I'm currently updating those. However, rst is not playing nice with the `|version|` placeholder when it is inlined in a numbered list and surrounded by the `.. warning` callout; it renders half of the `version`s correctly and half of them incorrectly. I'll see if I can resolve this, unless you are a Sphinx genius in which case please advise :)

eloquence: This is a big part of the reason why `update_version.sh` exists -- unfortunately we can't straightforwardly use `|version|` in inline snippets, so in those cases the script goes in and uses `sed` to perform a substitution. If you just add a similar substitution to the one already added for `backup_and_restore.rst` to the script, we should be golden.

docs/backup_and_restore.rst
```diff
@@ -101,89 +88,257 @@ When you are ready to begin the backup, run
...
+SSH (if configured) to their state at the time of the backup.
+
+If your backup contains v2 and v3 onion service configuration
+details, but your instance has since been reconfigured to use v3 onion
+services only, the restore command will restore only the v3 onion service
+configuration.
+
+.. note:: You cannot restore the onion service configuration from a v2-only
+  backup onto a v3-only server -  but you can restore submissions and account
+  data using the ``--preserve-tor-config`` argument. See :ref:`Data-only Restores <additional_restore_info>` for more information.
```
eloquence, line 157: The preceding two paragraphs have a lot of potential for being confusing, especially in light of the Ubuntu 20.04 migration.  Here's my stab at simplifying them, intended to replace both paragraphs with a single `.. note::`:

> You cannot restore v2 onion service configurations to a v3-only server (this includes any SecureDrop installation based on Ubuntu 20.04). If a restore operation would leave you without a valid onion service configuration, the restore script will fail with an error. In this case, you can still perform a [data-only restore using the `--preserve-tor-config` argument](linky).

Per https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/restore/files/compare_torrc.py I believe the restore script itself will print reasonably clear instructions, as well.

docs/includes/v3_backup_steps_common.rst
rocodes: fixed in 2e6a9b3

docs/includes/v3_backup_steps_common.rst
rocodes: fixed in 2e6a9b3

docs/includes/v3_backup_steps_common.rst
rocodes: fixed in 2e6a9b3

docs/includes/v3_backup_steps_common.rst
rocodes: This is the only change I didn't incorporate (a step called "Rename the backup file to..."). Rationale: there are instructions throughout that reference 'placeholder' filenames (`ossec.asc`, `journalist.asc`, `ca.crt`, etc), and I think if we say "we assume the file is called X" it's clear enough.

docs/includes/v3_backup_steps_common.rst
rocodes: fixed in 2e6a9b3

docs/backup_and_restore.rst
rocodes: Added this doc to `update_version.sh`, so all such should be resolved (thank you for this tip!)

docs/backup_and_restore.rst
```diff
@@ -101,89 +88,407 @@ When you are ready to begin the backup, run
...
 
-Once SSH access to the servers has been established (or if using SSH over
-local network), Onion URLs for the *Source Interface* and *Journalist Interfaces*
-can be fetched using the installer:
+#. Reinstall SecureDrop on the servers, following the :doc:`installation
+   instructions <install>`. During the configuration stage
+   (``./securedrop-admin sdconfig``), the values will be prepopulated based on
+   the old instance's configuration. Press **Enter** to accept each value,
+   except when you are asked if you want to enable v2 onion services - instead,
+   type ``no``.
```
eloquence, line 294: This section includes v3-only backups, in which case they won't have to type ``no``. I think that's reasonably obvious in context, but leaving a note in case you want to futz with the wording.

docs/backup_and_restore.rst
```diff
@@ -101,89 +88,407 @@ When you are ready to begin the backup, run
...
+   .. code:: sh
+
+      cd ~/Persistent/securedrop
+      export SD_OLD="~/Persistent/sd.bak/install_files/ansible-base"
+      export SD_NEW="~/Persistent/securedrop/install_files/ansible-base"
+      cp $SD_OLD/group_vars/all/site-specific $SD_NEW/group_vars/all/site-specific
+      cp $SD_OLD/sd-backup-old.tar.gz $SD_NEW/sd-backup-old.tar.gz
+
+   You will also need to copy the old instance's *Submission Public Key*,
+   *Ossec Alert Public Key*, and, if configured, the *Journalist Alert Public Key*.
```
eloquence, line 423: Ossec->OSSEC

update_version.sh
```diff
@@ -12,6 +12,7 @@ fi
 readonly OLD_VERSION=$(grep -oP '(?<=^release \= ")\d+\.\d+\.\d+' docs/conf.py)
 
 sed -i "s@$(echo "${OLD_VERSION}" | sed 's/\./\\./g')@$NEW_VERSION@g" docs/set_up_admin_tails.rst
+sed -i "s@$(echo "${OLD_VERSION}" | sed 's/\./\\./g')@$NEW_VERSION@g" docs/backup_and_restore.rst
```
eloquence, line 15: I can confirm that the version bump now works as expected. :)

docs/backup_and_restore.rst
```diff
@@ -101,89 +88,407 @@ When you are ready to begin the backup, run
...
+#. Copy the old instance's configuration files and backup from 
+   ``~/Persistent/sd.bak`` into ``~/Persistent/securedrop`` using the following 
+   Terminal commands:
+
+   .. code:: sh
+
+      cd ~/Persistent/securedrop
+      export SD_OLD="~/Persistent/sd.bak/install_files/ansible-base"
+      export SD_NEW="~/Persistent/securedrop/install_files/ansible-base"
+      cp $SD_OLD/group_vars/all/site-specific $SD_NEW/group_vars/all/site-specific
```
eloquence, line 259: In case folks type those commands by hand (and to make them easier to read), I would suggest  omitting the destination filename in all `cp` operations where it is not modified. So:

```bash
cp $SD_OLD/group_vars/all/site-specific $SD_NEW/group_vars/all/
cp $SD_OLD/tor_v3_keys.json $SD_NEW/
cp $SD_OLD/sd-backup-old.tar.gz $SD_NEW/
cp $SD_OLD/SecureDrop.asc $SD_NEW/
cp $SD_OLD/ossec.asc $SD_NEW/
cp $SD_OLD/journalist.asc $SD_NEW/
```

This is done further below for some of the commands.

eloquence:
Thanks @rocodes, this looks great to me. The comments I left are minor, and I'd be fine to merge as-is if needed. (We can tweak further during final QA as needed.)

docs/backup_and_restore.rst
rocodes: gotcha, fixed in  7f2b421

docs/backup_and_restore.rst
rocodes: addressed in  7f2b421

PR #172, opened 2021-03-12
Fix tilde expansion in instructions

eloquence:
## Status

Ready for review

## Description of Changes

The current Focal migration instructions put the `~` for user directory expansion in quotes. Unfortunately, tilde expansion [does not work inside double quotes](https://unix.stackexchange.com/questions/151850/why-doesnt-the-tilde-expand-inside-double-quotes). As a result the subsequent commands in the docs will fail as written.

The rest of the diff is automatic whitespace trimming.


conorsch (approved):
:+1:

PR #180, opened 2021-03-16
Update screenshots for static IP config in Tails

### LINKED ISSUE #173: Admin Workstation network config screenshot usage is confusing
The step-by-step instructions at https://docs.securedrop.org/en/stable/network_firewall.html#assign-a-static-ip-address-to-the-admin-workstation include helpful screenshots to configure the static IP address of the Admin Workstation, specifically showing the following two steps:

1) Change Addresses from Automatic (DHCP) to Manual (if it isn’t already).
2) Fill in the static networking information based on 3 NIC vs. 4 NIC configuration (two different screenshots)

The screenshot for 1) contains configuration details for the 3 NIC configuration, even though the only point of the screenshot is to illustrate how to enter the manual setup. This may cause the reader to apply the wrong config if they are not looking carefully, e.g., because they are trying to navigate quickly to the screenshot showing the required config.

Instead, we should show this screen in its default state before any values are entered, except for the Automatic->Manual change. Bonus points for highlighting the specific switch we're asking folks to flip.

## User Story

As an admin, I want screenshots to correspond to the steps they reference, so that I don't get disoriented during setup.

eloquence:
## Status
Ready for review

## Description of Changes

Fixes #173

- Clearer entrypoint screenshot that shows only the settings
  we're asking the user to change at that point in the process
- Update 4NIC screenshot per Tails 4 UI

`tails-manual-network.png` added in case we want to change the
highlight style at a later date.

## Checklist (Optional)

### DIFF
--- a/docs/images/firewall/IPv4_settings.png
+++ b/docs/images/firewall/IPv4_settings.png
@@ -1,3 +1,0 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:9466fb9c7a66d5daeca22e445cd293446dff62014ed7e59fb19437c2168755e3
-size 38102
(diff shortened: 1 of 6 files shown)


rocodes (approved):
LGTM, thanks for updating!

PR #181, opened 2021-03-16
Add docs for backing up logos

eloquence:
## Status
Ready for review. Whitespace clean up is in a separate commit,  beef2f0da18fa6276631112d862d4afe925a817f has the substantive changes.

## Description of Changes

Add instructions for backing up and restoring logos. These docs can be removed once a fix for freedomofpress/securedrop#5868 is included in a production release.

## Testing
Note that I've tested the `wget` v3 command (not the v2 one) on a prod Admin Workstation.

## Checklist (Optional)

### DIFF
--- a/docs/backup_and_restore.rst
+++ b/docs/backup_and_restore.rst
@@ -104,6 +104,24 @@ archive in the output of the backup command.
 
 .. include:: includes/backup-warning.txt
 
+.. _backing_up_logos:
+
+Backing up your custom logo
+'''''''''''''''''''''''''''
+
+The backup will not include any custom logo file that you have configured
+for your SecureDrop instance. To back up your logo:
+
+1. Open a terminal via **Applications ▸ System Tools ▸ Terminal**.
+2. Change into the persistent storage directory: ``cd ~/Persistent``
+3. Run the following command:
+
+.. code:: sh
+
+  wget $(grep Exec ~/source.desktop | awk '{print $2}')/org-logo -O logo-backup.png
+
+This will save your current logo file as ``logo-backup.png``.
+
 Restoring from a Backup
 -----------------------
 
@@ -189,7 +207,7 @@ Migrating Using a V2+V3 or V3-Only Backup
    and renamed ``sd-backup-old.tar.gz``.
 
 #. Move the existing *Admin Workstation* SecureDrop code out of the way, by
-   opening a Terminal via **Applications > System Tools > Terminal** and
+   opening a Terminal via **Applications ▸ System Tools ▸ Terminal** and
    running the command:
 
    .. code:: sh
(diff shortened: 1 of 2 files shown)


docs/backup_and_restore.rst
```diff
@@ -104,6 +104,30 @@ archive in the output of the backup command.
...
 
+.. _backing_up_logos:
+
+Backing up your custom logo
+'''''''''''''''''''''''''''
+
+As of SecureDrop |version|, the backup will not include any custom logo file
+that you may have configured for your SecureDrop instance. To back up the logo,
+download it using Tor Browser or on the command line. From ``~/Persistent``
+(a level above the ``securedrop`` directory), the following command will
```
eloquence, line 115: Suggested going to `~/Persistent` because this file is not included in `.gitignore`, so it would mess up future `git status` output. (https://github.com/freedomofpress/securedrop/issues/3502 can't come soon enough.)

docs/backup_and_restore.rst
```diff
@@ -104,6 +104,27 @@ archive in the output of the backup command.
 
 .. include:: includes/backup-warning.txt
 
+.. _backing_up_logos:
+
+Backing up your custom logo
+'''''''''''''''''''''''''''
+
+As of SecureDrop |version|, the backup will not include any custom logo file
```
zenmonkeykstop, line 112: This behaviour's been the same since the custom logo settings moved into the JI, by my casual read, so I think you can leave out the reference to the version (it might be misleading to folks who have an older backup and expect that it will get restored)

docs/backup_and_restore.rst
```diff
@@ -104,6 +104,27 @@ archive in the output of the backup command.
...
+that you may have configured for your SecureDrop instance. To back up the logo,
+download it using Tor Browser or on the command line. From ``~/Persistent``
+(a level above the ``securedrop`` directory), the following command will
```
zenmonkeykstop, line 115: It might be better to tell folks to open a terminal and explicitly include the command, ie:

Open a terminal via **Applications > System Tools > Terminal** and run the following commands:

```
cd ~/Persistent
wget ......
````
Also, if you want to remove the v2/v3 stuff, you could grab the canonical source address from the shortcut file! (just realized this!) with `cat ~/Desktop/source.desktop | grep Exec | awk '{print $2}'`

zenmonkeykstop (changes requested):
Looks good, some minor changes requested to remove ambiguity and make the instructions service-config-independent.

docs/backup_and_restore.rst
eloquence: Nice! Simplified a bit further and used in beef2f0da18fa6276631112d862d4afe925a817f

docs/backup_and_restore.rst
eloquence: Good point, removed in beef2f0da18fa6276631112d862d4afe925a817f, we'll just remove that language in its entirety once the issue is fixed.

zenmonkeykstop (approved):

PR #183, opened 2021-03-17
Additional migration details (SSH, extra admin USBS) 

### LINKED ISSUE #174: Migration docs should specify what to do with old `id_rsa` files
The migration docs at https://docs.securedrop.org/en/stable/backup_and_restore.html#migrate-v3 (and again for the v2 path) currently specify what to do with the `~/.ssh/config` and `~/.ssh/known_hosts` files, but not what to do with `id_rsa` and `id_rsa.pub`. Since those files will be overwritten if you follow the reinstall procedure to the letter, we should make the desired behavior explicit.

### LINKED ISSUE #175: Clarify network access requirements for migration procedure
In general I found the instructions in https://docs.securedrop.org/en/stable/backup_and_restore.html#migrating-using-a-v2-v3-or-v3-only-backup very easy to follow to migrate my long-running test instance to Ubuntu 20.04. :)

One thing I think we should spell out explicitly is that the *Admin Workstation* must be connected directly to the firewall. The *Admin Workstation* is explicitly designed to permit remote access over Tor, so we cannot assume that admins routinely keep it plugged in to the LAN port of the firewall -- they may in fact have not done so since an initial install years ago. 

I think we could add language to this effect at the beginning of the instructions, e.g.:

> During this procedure, your Admin Workstation must be directly connected to the network firewall. If this is not already the case:
> 1. Disable any active wireless connections in Tails.
> 2. Connect your *Admin Workstation* to the network firewall's LAN port, and configure [the *Admin Workstation's* static IP address in Tails](https://docs.securedrop.org/en/stable/network_firewall.html#assign-a-static-ip-address-to-the-admin-workstation). 
> 3. Wait for the "Tor is ready" notification to appear, which also confirms that your network connection is working.
>
> If you experience issues connecting to the network, carefully review both our public [firewall documentation](https://docs.securedrop.org/en/stable/network_firewall.html) and any notes specific to your setup, to ensure the network configuration matches the original one.

### LINKED ISSUE #176: Clarify post-migration instructions for preserving SSH access on all Admin Workstations
In all migration docs, we should clarify the procedure for preserving SSH access from additional Admin Workstations beyond the one that is used to perform the migration. 

Depending on resolution for #174, this may involve copying the SSH keys; it certainly involves copying the Monitor Server's new onion address and v3 keys (its original config is not preserved).

The [Focal Migration docs](https://docs.securedrop.org/en/stable/upgrade/focal_migration.html) currently [point to](https://github.com/freedomofpress/securedrop-docs/blame/main/docs/upgrade/focal_migration.rst#L242-L246) https://docs.securedrop.org/en/stable/v3_services.html#update-tails-v3; however, this is written from the point of view of ensuring v3 access after v2 is disabled, and it's not clear that it's a required step even for v3-only instances.

## User Story

As an admin, I don't want to be caught by surprise after a migration that I can access `app` but not `mon` from one of my Admin Workstations, so that I can perform maintenance tasks without issues.

rocodes:
## Status
Ready for review 

## Description of Changes

* Description:  Add instructions for repairing additional Admin USBs. Add notes on rotating SSH key during migration.

* Fixes #174
* Fixes #175
* Fixes #176 

## Testing
* Visual review
* Step through mon ssh repair 

## Release 
* n/a 

## Checklist (Optional)

### DIFF
--- a/docs/backup_and_restore.rst
+++ b/docs/backup_and_restore.rst
@@ -215,12 +215,16 @@ Migrating Using a V2+V3 or V3-Only Backup
       mv ~/Persistent/securedrop ~/Persistent/sd.bak
 
 #. Move the existing *Admin Workstation* SSH configuration out of the way via
-   the Terminal, using the commands:
+   the Terminal, using the command:
 
    .. code:: sh
 
-       mv ~/.ssh/config ~/.ssh/config.bak
-       mv ~/.ssh/known_hosts ~/.ssh/known_hosts.bak
+      find ~/.ssh/ -type f -exec mv {} {}.bak \;
+
+   .. note::
+      You will be generating fresh SSH credentials for the servers, and any
+      other *Admin Workstation* USBs will have to be
+      :ref:`provisioned with updated credentials <repair_admin_usbs>`.
 
 #. Re-clone the SecureDrop repository to the *Admin Workstation* using the following
    Terminal commands:
@@ -299,9 +303,9 @@ Migrating Using a V2+V3 or V3-Only Backup
       cp $SD_OLD/sd.{crt,key} $SD_NEW/
       cp $SD_OLD/ca.crt $SD_NEW/
 
-#. If you are migrating to new hardware, ensure your old servers have been
-   decommissioned and/or destroyed by following the relevant sections of
-   :doc:`our decommissioning documentation <decommission>`.
+#. Ensure your *Admin Workstation* is connected to a LAN port on your
+   network firewall, and
+   :ref:`configure the Admin Workstation's IP address <assign_static_ip_to_workstation>`.
 
 #. Install Ubuntu 20.04 on the *Application* and *Monitor Servers*, following
    the :doc:`server setup instructions<servers>` to install with the correct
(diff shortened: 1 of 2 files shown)


docs/backup_and_restore.rst
```diff
@@ -197,12 +197,16 @@ Migrating Using a V2+V3 or V3-Only Backup
...
+   the Terminal, using the command:
 
    .. code:: sh
 
-       mv ~/.ssh/config ~/.ssh/config.bak
-       mv ~/.ssh/known_hosts ~/.ssh/known_hosts.bak
+      find ~/.ssh/ -type f -exec mv {} {}.bak \;
+
+   .. note::
+      You will be generatating fresh SSH credentials for the servers, and any
```
eloquence, line 207: typo: generatating->generating

docs/backup_and_restore.rst
```diff
@@ -334,6 +338,74 @@ Migrating Using a V2+V3 or V3-Only Backup
...
+#. Copy the following files from your primary *Admin Workstation* onto the
+   Transfer Device:
+   
+   - ``~/Persistent/securedrop/install_files/ansible-base/tor_v3_keys.json``
+   - ``~/Persistent/securedrop/install_files/ansible-base/mon-ssh.auth_private``
+   - ``~/.ssh/id_rsa.pub``
+   - ``~/.ssh/id_rsa``
+
+#. Boot into each additional Admin Workstation, setting an administrative 
+   passphrase during startup. Once logged in, attach the Transfer Device
```
eloquence, line 365: I would say "set [an administration password](https://tails.boum.org/doc/first_steps/welcome_screen/administration_password/) and unlock the persistent volume on the Tails welcome screen" (that's more consistent with the language we've used elsewhere in the docs).

docs/backup_and_restore.rst
```diff
@@ -334,6 +338,74 @@ Migrating Using a V2+V3 or V3-Only Backup
...
+
+      cd ~/Persistent/securedrop
+      ./securedrop-admin tailsconfig
+
+#. Test connectivity to each server by running ``ssh app uptime`` 
+   and ``ssh mon uptime``.       
+
+#. Once all *Admin Workstations* have been updated, securely wipe the files on 
+   the Transfer Device, by right-clicking them in the file manager and selecting 
+   **Wipe**. Do not skip this step. Then, reformat the Transfer device using the
```
eloquence, line 397: I would remove "Do not skip this step". If you want the Wipe/Format steps to stand out clearly, we could separate them.

docs/backup_and_restore.rst
```diff
@@ -334,6 +338,74 @@ Migrating Using a V2+V3 or V3-Only Backup
...
+   **Disks** utility.   
+
+   .. note::
+      Alternatively, if you wish to use different SSH credentials for each
+      *Admin Workstation*, you may do so. In this case, use a clean Transfer
+      Device to transfer the public portions of those keys to your primary 
+      *Admin Workstation*, where you will add them to the servers' 
+      ``authorized_keys`` files, as described :ref:`here <ssh_add_pubkey>`. 
+      You may also `contact Support`_ for assistance.
```
eloquence, line 406: - This feels like it might be better placed in context, or in the preface of this section.
- In any event, I would suggest making it clearer that this only pertains to the `id_rsa`/`id_rsa.pub` portion of the instructions -- they'll still need to copy the service details to each admin stick.

docs/backup_and_restore.rst
```diff
@@ -334,6 +338,74 @@ Migrating Using a V2+V3 or V3-Only Backup
...
+   - ``~/Persistent/securedrop/install_files/ansible-base/mon-ssh.auth_private``
+   - ``~/.ssh/id_rsa.pub``
+   - ``~/.ssh/id_rsa``
+
+#. Boot into each additional Admin Workstation, setting an administrative 
+   passphrase during startup. Once logged in, attach the Transfer Device
+   and unlock it.
+
+#. Ensure that this Admin Workstation is using an up-to-date version of Tails
+   and is running the latest SecureDrop Application code, |version|.
```
eloquence, line 369: Application->application

docs/backup_and_restore.rst
```diff
@@ -334,6 +338,74 @@ Migrating Using a V2+V3 or V3-Only Backup
 #. :doc:`Test the new instance <test_the_installation>` to verify that the
    web interfaces are available and the servers can be reached via SSH.
 
+.. _repair_admin_usbs:
+
+Repair Additional Admin Workstations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
eloquence, line 344: I would suggest a standard scrub of the below for italicizing glossary terms (I don't have strong attachments to the convention, but while we have it ..).

eloquence:
Thanks @rocodes! At a high level, the process makes sense to me; my comments are mainly smaller nits. I've not tested the full procedure yet, but I have verified that the `find` command works as expected. If you have done a full procedural test, I would consider that sufficient to merge.

I'll admit that I personally just did the *plug Admin Workstation 2 into Admin Workstation 1 and copy files over* thing. There are reasonable security arguments to use an additional device, but also some additional opsec risks. As we do real-world migrations with admins, I would suggest we keep an eye out for these complexity traps, where real world procedures may not end up reflecting what we put in our docs.

rocodes:
> I'll admit that I personally just did the _plug Admin Workstation 2 into Admin Workstation 1 and copy files over_ thing. There are reasonable security arguments to use an additional device, but also some additional opsec risks. As we do real-world migrations with admins, I would suggest we keep an eye out for these complexity traps, where real world procedures may not end up reflecting what we put in our docs.

These are reasonable concerns. I'll defer to others on what the best procedure is, but my reasons for using an intermediate device are: 
- potentially error-prone and/or destructive operation with the only good Admin Workstation; best principle is to propose the nondestructive path
- people who can understand these instructions and come up with a shortcut are welcome to do so

docs/backup_and_restore.rst
rocodes: sounds good. I find all the italics and capitalization of names really hard to read, personally, so when it says *Admin Workstation* and *Transfer Device* everywhere I find it confusing. I've changed to "LUKS-encrypted USB" here (fc894a2)

docs/backup_and_restore.rst
rocodes: fixed in fc894a2

rocodes:
Thanks for your comments, addressed in fc894a2

eloquence:
Thanks @rocodes! :) Is this ready for final review, or do you want to pull in #175 as well?

rocodes:
Ok, updated to also address #175. If we're happy with the 'luks usb to copy credentials to other admin sticks' approach, this is ready for review+merge.

eloquence:
LGTM; I pushed one small commit for whitespace/formatting tweaks and a typo (encryped->encrypted). We do have a merge conflict with the changes from #181. In case it's helpful for your copy/paste convenience, I've pushed 3782d8ba5bd7a25e761feecb7b0a4ef6f92224b2 in a separate branch, which squashes the changes in this PR into one commit, and resolves the conflict with `main` (I didn't want to force-push to this branch to avoid messing with your commit authorship).

eloquence (approved):
Looks great, thanks again for your work on this @rocodes!

PR #189, opened 2021-03-26
Bump all docs requirements

eloquence:
## Status

Ready for review

## Description of Changes

Bumps requirements to latest to ensure we pick up bugfixes, enhancements, security and performance fixes for all dependencies. I think we can avoid being especially selective here -- tested on Python 3.6 and 3.7 virtualenvs.

Command used to bump requirements:
`pip-compile --allow-unsafe --generate-hashes requirements.in --output-file requirements.txt`

The linkcheck appears to have been improved and picked up a few previously unnoticed issues, which are also addressed by this PR.

## Testing
Create a virtulaenv with a supported version of Python and check that `make` targets are still working as expected.

## Checklist (Optional)


zenmonkeykstop (approved):
made a Python 3.7 vitrualenv, everything checks out (and linkcheck seems a lot faster which is great!)

Approved.

PR #141, opened 2021-01-27
Re-organize translation docs by Divio's system

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

This draft PR explores re-organizing the translation docs [with the Divio system in mind](https://github.com/freedomofpress/securedrop-docs/issues/134#issuecomment-768279483). The goal wasn't to achieve a production-ready slice of documentation, but to draft enough of it to be able to evaluate if it's a direction worth pursuing. **Edit to add**: the answer to the first iteration was "worth pursuing", the goal of the second iteration is to start checking how all pieces can fit together. **Edit 2**: we're now refining the content within the new sections.

**Worth looking at**

- how the sidebar navigation looks (as an overview of the section)

Fixes #134 (maybe!)

## Questions / goals:

-  Would consolidating both the i18n and l10n pages in this way make this content easier to consume for the different people who need it?
-  Would that also make it is easier to extend and maintain?

## Checklist (Optional)

- [x] Doc linting (`make docs-lint`) passed locally
- [x] Doc link linting (`make docs-linkcheck`) passed
- [x] You have previewed (`make docs`) docs at http://localhost:8000
- [ ] The commits are squashed as intended (no `fixup` in commit messages)

Issue #70, opened 2020-10-20
Audit image labels for accessibility


mrphs:
I like how detailed SecureDrop docs are, and it seems like you've done your best not to depend on screenshots, however they are absolutely useful and I just realized they don't seem to have the "title" attribute. This is often overlooked as unnecessary but when I was doing UX studies some time ago, I learned that's something visually impaired people highly depend on.

I know that you're working on new home for documentations. Perhaps this can be something to think about implementing while making the new website. :)

eloquence:
@joaedwar Good meaty candidate issue for you to work on, will require reviewing all image labels to make sure they do in fact help make our documentation more accessible, e.g., when read out loud to a visually impaired person via a screen reader.

PR #196, opened 2021-04-10
Improve integration of images with surounding text using the figure element

gonzalo-bulnes:
## Status

Work in progress

## Description of Changes

Adds captions to all images in the documentation by using the [`figure`][fig] directive. Each caption describes the image and hints that clicking on the image will display it in full size. Since the possibility of opening the image in full size is now explicit, this PR aslo slightly reduces the size of the images in the docs in order to reduce the interruption that illustrations cause to the flow of the text.

  [fig]: https://docutils.sourceforge.io/docs/ref/rst/directives.html#figure

This is a follow up on the introduction of image captions in #141.

### Pros

- Image descriptions make content more accessible.
- Images don't interrupt the text flow as much.
- It is clear that images can be opened full-size.

### Accepted trade-offs

- Unlike the `image` directive, `figure` cannot be used in [`substitution definitions`][subst], which means that the image URLs cannot be grouped at the end of the document anymore. Because most images are used only once per document, I think that's an acceptable trade-off for the accessibility improvement.

  [subst]: https://docutils.sourceforge.io/docs/ref/rst/directives.html#directives-for-substitution-definitions

## Testing

_I'll update when the PR is ready for review._

## Release 

_No special considerations before release come to mind yet, I'll confirm when the PR is ready for review._

## Example

image

Note: the pixellation on the image is due to the amount of zoom out I applied to keep the screenshot size small. The figures don't appear pixellated in the docs ([example](https://docs.securedrop.org/en/latest/development/translations.html)).

## Checklist (Optional)

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -216,7 +216,11 @@ passphrase, and two-factor authentication token to log in.
 
 By default, you will be logged in to the *Journalist Interface*'s source list page.
 
-|SecureDrop main page|
+.. figure:: images/manual/screenshots/journalist-admin_index_no_documents.png
+   :align: center
+   :alt: The top navigation of the Journalist Interface says 'Logged on as Journalist' and displays an 'Admin' link.
+   :figwidth: 80%
+   :width: 100%
 
 In the course of normal administration operations you should not need to view source
 communications, but if you do, you can find information on managing submissions in
@@ -254,11 +258,19 @@ account setup.
 
 #. Click **Admin** in the top right corner of the page to load the *Admin Interface*.
 
-   |SecureDrop admin home|
+   .. figure:: images/manual/screenshots/journalist-admin_interface_index.png
+      :align: center
+      :alt: The Admin Interface displays an 'Add User' button.
+      :figwidth: 80%
+      :width: 100%
 
 #. Click **Add User** to add a new user.
 
-   |Add a new user|
+   .. figure:: images/manual/screenshots/journalist-admin_add_user_totp.png
+      :align: center
+      :alt: The form used to create new users displays a pre-generated Diceware passphrase.
+      :figwidth: 80%
+      :width: 100%
 
 #. Hand the keyboard over to the journalist so they can create their own username.
 #. Once they’re done entering a username for themselves, have them save their pre-generated Diceware passphrase to their password manager.
@@ -275,7 +287,13 @@ FreeOTP
 
 #. If the journalist is using FreeOTP or another app for two-factor authentication, click **Add User** to proceed to the next page.
 
-   |Enable FreeOTP|
+   .. figure:: images/manual/screenshots/journalist-admin_new_user_two_factor_totp.png
+      :align: center
+      :alt: The form used to enable FreeOTP displays a barcode and a two-factor secret.
+      :figwidth: 80%
+      :width: 100%
+
+      Click on the image to see it full-size.
 
 #. Next, the journalist should open FreeOTP on their smartphone and scan the barcode displayed on the screen.
 #. If they have difficulty scanning the barcode, they can tap on the icon at the top that shows a plus and the symbol of a key and use their phone's keyboard to input the two-factor secret into the ``Secret`` input field, without whitespace.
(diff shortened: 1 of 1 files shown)


gonzalo-bulnes:
Touching base to check if this is desirable.

It's a quick run through the docs for me, so I think it's worth rolling out captions everywhere (and give a consistent pattern to follow for new docs). I'm also taking note of small UX inconsistencies as I compare the docs and the screenshots so the read-through doesn't feel like wasted time for me.

That being said YMMV, hence the question before I go further :wink:

docs/admin.rst
```diff
@@ -333,7 +373,7 @@ Setting Submission Preferences
 By default, SecureDrop supports both text submissions and document uploads. If you
 only want to receive text messages, you can disable uploads as follows:
 
-* check the the **Prevent sources from uploading documents** checkbox
+* check the **Prevent sources from uploading documents** checkbox
```
eloquence, line 376: nice find, someone should make a linter that just catches `the the` repetitions ;-)

eloquence:
@gonzalo-bulnes Thank you for working on this! I'm broadly in favor of this change, but I think if we use captions throughout the docs like this, we may want to style them a little bit differently, e.g., a bit smaller and in dark grey font, so they don't interrupt the flow of the main body text.

I'm curious what other maintainers and contributors think about this change. An alternative approach would be to be much more selective, and only apply it to graphics where the detail really is important (like the overview diagram, and the detailed firewall config screenshots in https://docs.securedrop.org/en/stable/network_firewall.html).

docs/admin.rst
gonzalo-bulnes: :P

gonzalo-bulnes:
I tend not to question the styles much when relying on a theme like Read The Docs'. I guess I tend to assume that readability being their core business, the themes would have reasonable defaults, but yours is a fair point. Getting people's thoughts seems like a good idea! :slightly_smiling_face:

gonzalo-bulnes:
:wave: As I was [taking a look at the Netgate docs a few days ago](https://gist.github.com/gonzalo-bulnes/3b23d6af0949b36f607dd898337d6471#netgate-step-by-step-instructions), I noticed the way they use captions when including figures in their how-tos. Specifically, they use them to repeat the action you're supposed to perform when your see what the image show.
Today while reading your comments again, I just made the link and I wonder what you'd think @eloquence : )  

Example in the Netgate docs: ([source](https://docs.netgate.com/pfsense/en/latest/solutions/sg-3100/initial-configuration.html))

example-of-figure-captions-in-netgate-docs

gonzalo-bulnes:
Not directly related:

> An alternative approach would be to be much more selective, and only apply it to graphics where the detail really is important [...]

When it comes to including "Click on the image to see it full-size", I agree it's probably not useful to mention it explicitly in all images. Besides, the feature doesn't depend on the caption itself.

eloquence:
@gonzalo-bulnes I like the Netgate example. The screenshots have arrow annotations, and the captions restate the specific step that users are asked to perform.

One thing I noticed is that your PR implicitly changes the `ALT` attribute for screenshots. For example, `journalist-admin_interface_index.png` currently has the moderately useful `ALT` attribute "SecureDrop admin home" in https://docs.securedrop.org/en/stable/admin.html

After your change, that `ALT` attribute is replaced with "_images/journalist-admin_add_user_totp.png". I believe this is because you are removing the image labels as part of transitioning to the `figure` syntax.

I would suggest the following design goals:
- We should aim for `ALT` attributes that describe the contents of the screenshot that is being depicted. See discussion in #70.
- Captions should be included if/when they are useful to restate a specific instructional step, or to explain more complex images.
- Navigational text like "Click to enlarge" should be styled to be unobtrusive or used selectively.

Do those goals make sense to you? If so, do you want to take another stab at a selective edit e.g. to the Admin Guide in line iwth these design goals?

gonzalo-bulnes:
Yes @eloquence , all three goals make complete sense to me! I'm happy to use the Admin Guide as a sample 👍

I hadn't noticed the change in the `alt` attributes, thank you for pointing it out. I'll have #70 in mind while I do this.

gonzalo-bulnes:
I kept the "Click to enlarge" hints when the screenshot displays large amount of text. It's admittedly an arbitrary threshold, and I'm not 100% sold on the idea.

On one side I wouldn't mind removing the hints entirely, on the other I think it's good to give people a change to learn that images can be zoomed - and that seems more relevant when there screenshot displays a large amount of information.
What do you think @eloquence?

eloquence:
Apologies for not responding sooner, @gonzalo-bulnes, thanks for the follow-up. I would not add the "Click to see full-size" hint to the larger SecureDrop screenshots; as you say, the size threshold is somewhat arbitrary, which could be confusing for documentation writers and readers alike.

I think the resize hint would be helpful in these cases:
- The overview diagram in https://docs.securedrop.org/en/stable/overview.html
- The architecture diagram in https://docs.securedrop.org/en/stable/development/tips_and_tricks.html#architecture-diagrams
- The data flow diagram in https://docs.securedrop.org/en/stable/threat_model/dataflow.html

The firewall screenshots in https://docs.securedrop.org/en/stable/network_firewall.html may also benefit from the hint. But I would err on the size of conservatism for now; perhaps we can come up with a more subtle affordance to make the resizability of images discoverable, separately from this PR.

gonzalo-bulnes:
Thanks @eloquence I'll apply that to see how it looks/reads 🙂

gonzalo-bulnes:
(rebased, squashed commits and removed the hints)

gonzalo-bulnes:
Arguably, this PR is now more about using the `figure` element where useful, more than adding captions (the cases where captions are useful are sparse). Do you mind me renaming the PR @eloquence? I'm thinking about something in the lines of impove the images integration in the text, or specifically employing the `figure` element for improved readability.

With the focus on using the `figure` element, I could add the other pages to this same PR: those that could use captions (e.g. the diagrams you mentioned above) and where the flow of reading could use the narrower images. How does that sound to you?

eloquence:
Sounds great, please retitle as you see fit :)

eloquence:
@gonzalo-bulnes Just checking in on the status of this PR, would be great to get a first iteration over the finish line in the next couple of weeks :)

gonzalo-bulnes:
Hey, apologies for the slow reply, I plan on resuming this tomorrow morning : )

eloquence:
@gonzalo-bulnes If you're coming up from air sometime, this may be a nice slow-burn PR to return to :P

eloquence:
Hi @gonzalo-bulnes, this PR has been stale for a while, I'd suggest closing it out for now unless you want to return to it.

gonzalo-bulnes:
I'm properly tracking this one now in my personal project board. :grimacing:

eloquence:
Hi @gonzalo-bulnes, I suggest we close this out or adopt it, but let me know if you have interest in picking it up again :)

gonzalo-bulnes:
I won't be resuming this in the near future, I'll close it an you can re-open whenever you see fit 🙂

Issue #184, opened 2021-03-18
Improve Landing Page docs


rocodes:
## Describe the change
The landing page guidelines are not just guidelines--there are certain requirements we have for listing an instance in our directory. We should clarify this and provide an easy way for instances to tell if they are in compliance or not (e.g. share our pass/fail criteria).

Also, while we're in there, we could consider updating our criteria to include the `Feature-Policy` (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy) aka `Permissions-Policy` header (set to `none`). This is not supported by all browsers, but controls what kind of access the browser has to microphone, geolocation, battery status, and other APIs. 

## How will this impact users?
Landing page "guidelines" will be clearer and easier to interpret/implement before submitting instance verification form

## User Research Evidence
n/a just lots of instance verification correspondence

gonzalo-bulnes:
Hi @rocodes, if you have a list of the requirements that an instance's landing page must meet to be listed in the directory, I think that would be useful to get an idea of how much work this page needs. I'm thinking it may be "not too much", but I may well be very wrong :slightly_smiling_face: 

Related issues: https://github.com/freedomofpress/securedrop-docs/issues/37, https://github.com/freedomofpress/securedrop-docs/issues/55

gonzalo-bulnes:
Also related: https://github.com/freedomofpress/securedrop-docs/issues/46

gonzalo-bulnes:
Related: https://github.com/freedomofpress/securedrop-docs/issues/210 (A place should be found in the landing page docs for that new information when it's defined.)

PR #198, opened 2021-04-11
Add Landing Page testing recommendation

### LINKED ISSUE #39: Landing page guidelines should recommend testing in Tor
The [landing page guidelines](https://docs.securedrop.org/en/stable/deployment/landing_page.html) should strongly recommend testing a landing page in Tor before advertising it. Sources may arrive on the landing page using Tor, and many sites are configured with security measures that only kick in when Tor is in use (warnings specific to certain exit nodes, captchas, DDoS protections, etc.). It's a very bad experience for a source if visiting a landing page in Tor doesn't work as expected.

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

Adds a strong recommendation to test a newly deployed *Landing Page* using Tor, along with some explanation of why that's important.

The boundary between explanations and instructions is currently unclear in this page (see https://github.com/freedomofpress/securedrop-docs/issues/184). I feel like some explanation were required, but also that concrete instructions would be helpful. Instead of creating two separate sections that would fit awkwardly in the page, I decided to format this new section as a mix of explanation and how-to which can be revisited when #184 is addressed. Thoughts welcome!

Fixes #39 

## Testing

- [ ] Proof-read please :slightly_smiling_face: 
- [ ] Check that the Tor Project link is correct

## Release 

_No special considerations for release._

## Checklist (Optional)

### DIFF
--- a/docs/deployment/landing_page.rst
+++ b/docs/deployment/landing_page.rst
@@ -308,6 +308,42 @@ the attack surface.
 .. _`fail2ban` : https://github.com/fail2ban/fail2ban
 .. _`sshguard` : https://bitbucket.org/sshguard/sshguard/
 
+How to test your *Landing Page* using Tor Browser
+-------------------------------------------------
+
+*Sources* may visit your *Landing Page* using Tor.
+
+Many websites are configured with security measures that only apply
+when Tor is in use. For example, Tor visitors may be requested to solve
+a CAPTCHA, may trigger warnings that are specific to some Tor exit nodes,
+or may be unable to load the page altogether because of
+Tor-specific DDoS protections.
+
+The effect of such measures cannot be tested without using Tor, and it is
+a very bad experience for a *source* if visiting a *Landing Page* doesn't work
+as expected. Because of that, we **recommended strongly** that you test
+your organization's *Landing Page* using Tor *before* you start advertising it.
+
+You can do it using Tor Browser:
+
+#. Download Tor Browser from the `Tor Project website`_.
+#. Visit your *Landing Page*.
+#. Ensure the `Tor Browser security level`_ is set to "Safest"
+   by clicking on the shield icon. Click on "Advanced Security Settings"
+   and select "Safest" if necessary.
+#. Verify that everything works as expected
+
+Repeat the test several times after reloading the page `using a different Tor circuit`_ each time.
+
+#. Reload the page using a new Tor circuit by clicking on "New Tor Circuit for this Site"
+   in the site information menu (padlock icon in the URL bar) or in the hamburger menu.
+#. Verify that everything still works as expected
+   when the Tor Browser security level is set to "Safest".
+
+.. _`Tor Project website`: https://www.torproject.org/
+.. _`Tor Browser security level`: https://tb-manual.torproject.org/security-settings/
+.. _`using a different Tor circuit`: https://tb-manual.torproject.org/managing-identities/
+
 *Landing Page* Content Suggestions
 ----------------------------------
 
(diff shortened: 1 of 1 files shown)


docs/deployment/landing_page.rst
```diff
@@ -384,3 +384,27 @@ SecureDrop submissions. We control the servers that store your submissions,
...
+The effect of such measures cannot be tested without using Tor, and it is
+a very bad experience for a *source* if visiting a *Landing Page* doesn't work
+as expected. Because of that, it is **strongly recommended** that you test
+your organization's *Landing Page* using Tor *before* you start advertising it.
+
+You can do it using Tor Browser:
+
+#. Download Tor Browser from the `Tor Project website`_.
+#. Visit your *Landing Page*.
+#. Verify that everything works as expected.
```
eloquence, line 408: I would add a point here about reloading the page using different Tor circuits to observe behavior across a range of exit nodes.

This can be done by clicking the padlock icon on the left side of the address bar, and clicking "New Circuit for this Site".

docs/deployment/landing_page.rst
```diff
@@ -384,3 +384,27 @@ SecureDrop submissions. We control the servers that store your submissions,
...
+
+Many websites are configured with security measures that only apply
+when Tor is in use. For example, Tor visitors may be requested to solve
+a CAPTCHA, may trigger warnings that are specific to some Tor exit nodes,
+or may be unable to load the page altogether because of
+Tor-specific DDoS protections.
+
+The effect of such measures cannot be tested without using Tor, and it is
+a very bad experience for a *source* if visiting a *Landing Page* doesn't work
+as expected. Because of that, it is **strongly recommended** that you test
```
eloquence, line 401: I would change this to active voice (it is recommended -> we recommend).

docs/deployment/landing_page.rst
eloquence: Another thing worth testing is if the page loads correctly if the Tor Browser security settings are set to "Safest".

docs/deployment/landing_page.rst
```diff
@@ -384,3 +384,27 @@ SecureDrop submissions. We control the servers that store your submissions,
 so no third party has direct access to the metadata or content of what you send us.
 
 Do not discuss leaking or whistleblowing, even with trusted contacts.
+
+How to test your *Landing Page* using Tor Browser
```
eloquence, line 388: I agree with you that this whole section could use an information architectural overhaul; as it is, I would prefer to place this new section before the sample landing page content. It is a consideration that should be kept in mind during development and deployment of the landing page, anyway, as it often will require specific configuration changes to the CDN or the webserver to prevent Tor-specific issues.

eloquence (changes requested):
Thanks @gonzalo-bulnes, a few small suggestions but otherwise this looks great!

docs/deployment/landing_page.rst
gonzalo-bulnes: I had a go at both, let me know if that's what you had in mind. : )

docs/deployment/landing_page.rst
gonzalo-bulnes: Done.

docs/deployment/landing_page.rst
gonzalo-bulnes: Yep, that makes sense. I moved it just above the content recommendations.

gonzalo-bulnes:
Because it is a short section, I applied your suggestions (and rebased the branch) at once.

I added two links to the relevant Tor Browser documentation for setting the security level and reloading a site using a new Tor circuit. Both pages give a bit of background and explain the different ways to do both. They seem short and useful to me. What do you think @eloquence?

eloquence (approved):
LGTM, tweaked/consolidated slightly. Thank you!

gonzalo-bulnes:
Yes, the single list indeed reads better. 👍

PR #201, opened 2021-04-14
Clarify SSH wait time post install

rocodes:
## Status
Ready for review

## Description of Changes

* Description:  add note about time delay for SSH accessibility post-install. Also clarifies wording of the server reinstall to include the `install` command.

## Testing
* visual review

## Release 
* n/a 

## Checklist (Optional)

### DIFF
--- a/docs/backup_and_restore.rst
+++ b/docs/backup_and_restore.rst
@@ -300,6 +300,11 @@ Migrating Using a V2+V3 or V3-Only Backup
    settings, test connectivity, and set up SSH keys to allow for
    *Admin Workstation* access.
 
+   .. note::
+
+      You may need to wait approximately 10-15 minutes after installing 
+      Ubuntu 20.04 for the servers to become reachable via SSH.
+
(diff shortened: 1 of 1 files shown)


docs/backup_and_restore.rst
```diff
@@ -521,20 +527,28 @@ source accounts, and journalist accounts. To do so, follow the steps below:
    settings, test connectivity, and set up SSH keys to allow for
    *Admin Workstation* access.
 
+   .. note::
+
+      You will need to wait approximately 10-15 minutes after installing
```
rmol, line 532: We could say, "You _may_ need to wait ..." instead of adding a mandatory extra 20 minutes into the process.

rmol (changes requested):
I think it's fine. If not every migration so far has encountered the delay, we could suggest that they might need to wait, but if the servers are reachable, they should feel free to proceed.

rmol (approved):

Issue #213, opened 2021-05-03
Remove v3 onion service docs


eloquence:
#208 removed references to v2 onion services, but we still have a dedicated section focused on configuring/enabling v3 onion services:

https://docs.securedrop.org/en/latest/v3_services.html

This section can be safely removed, but for orgs who have missed the Ubuntu 20.04 migration deadline, we will need to integrate the information about [migrating workstation access to v3](https://docs.securedrop.org/en/latest/v3_services.html#update-tails-v3) into the migration guide if we do so.

gonzalo-bulnes:
I suppose that whether or not some organizations are still using Ubuntu 16.04 and v2 onion services is probably sensitive. However, when reviewing the docs over the past two days (https://github.com/freedomofpress/securedrop-docs/issues/236) I wondered if the deprecation warning and special notices should be removed. Whenever that's the case I'm happy to do it, or provide some context and list the changes as a "good first issue" for someone else to pick.

PR #214, opened 2021-05-04
Removes Xenial-specific references from the docs

conorsch:
## Status

Ready for review 

## Description of Changes

Follow-up to https://github.com/freedomofpress/securedrop/pull/5911,
in which we removed Xenial-specific codepaths from the code repo.
Here, we do the same throughout the documentation. There are notable
exceptions:

  * https://docs.securedrop.org/en/stable/v3_services.html
  * https://docs.securedrop.org/en/stable/upgrade/focal_migration.html

Those URLs remain untouched by these changes. We can follow up soon with
removal for those, too, but it seems helpful to maintain a bit longer.

## Testing
1. Check for any missed references to the old OS. Try `grep -ri xenial` and `grep -r 16.04`, for instance. 
2. Do you agree we should preserve the v3 services & focal migration docs for now? 

## Release 
Intentionally deferring a few docs changes (see above), to preserve URLs a while longer. If folks find that too conservative, happy to remove in this PR. 

## Checklist (Optional)

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -437,15 +437,6 @@ VERSION_CODENAME is "focal"
 
   sudo unattended-upgrades
 
-
-VERSION_CODENAME is "xenial"
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-.. code:: sh
-
-  sudo cron-apt -i -s
-
-
 .. note::
 
    Depending on the nature of the update (e.g., if the ``tor`` package is
--- a/docs/development/qubes_staging.rst
+++ b/docs/development/qubes_staging.rst
@@ -1,9 +1,9 @@
 Virtual Environments: Using Qubes
 =================================
 
-SecureDrop currently uses Ubuntu Xenial as its server OS, and Focal support is under
-development. The instructions below cover setting up a SecureDrop staging environment
-using either Xenial or Focal under Qubes.
+SecureDrop currently uses Ubuntu Focal as its server OS.
+The instructions below cover setting up a SecureDrop staging environment
+using Focal under Qubes.
 
 It is assumed that you have an up-to-date Qubes installation on a compatible
 laptop, with at least 16GB RAM and 60GB free disk space. The SecureDrop server VMs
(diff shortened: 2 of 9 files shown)


docs/development/qubes_staging.rst
```diff
@@ -245,8 +239,9 @@ to set up the development environment.
 
 Once finished, build the Debian packages for installation on the staging VMs:
 
-- **Xenial:** use the command ``make build-debs``
-- **Focal:** use the command ``make build-debs-focal``
+.. code::  sh
+
+    make build-debs-focal
```
eloquence, line 244: That Makefile target does not exist, it's just build-debs

conorsch: Thank you, fixed!

docs/admin.rst
```diff
@@ -437,15 +437,6 @@ VERSION_CODENAME is "focal"
...
 
-
-VERSION_CODENAME is "xenial"
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-.. code:: sh
-
-  sudo cron-apt -i -s
-
-
```
eloquence, line 448: I think by 2.10 or 2.2.0 we can remove the OS platform check here from the docs entirely, but fine to keep for now to avoid any confusion.

docs/development/testing_continuous_integration.rst
```diff
@@ -117,48 +116,48 @@ of a commit to apply to a branch in order disable the deletion for the Focal sta
...
-    -
-         - store_test_results:
-             path: ~/sd/junit
-
-    diff --git a/devops/gce-nested/ci-go.sh b/devops/gce-nested/ci-go.sh
-    index 850324ecc..776120df4 100755
-    --- a/devops/gce-nested/ci-go.sh
-    +++ b/devops/gce-nested/ci-go.sh
-    @@ -16,4 +16,3 @@ export BASE_OS="${BASE_OS:-xenial}"
+   diff --git a/.circleci/config.yml b/.circleci/config.yml
```
eloquence, line 119: Tested that diff applies cleanly on current `develop`. A bit iffy to maintain this way in docs though, maybe we can script that in future?

eloquence (approved):
> Check for any missed references to the old OS. Try grep -ri xenial and grep -r 16.04, for instance.

All LGTM there!

>  Do you agree we should preserve the v3 services & focal migration docs for now?

I think we can ditch the v3 docs soon, but that's tracked in #213.

PR #218, opened 2021-05-12
make build-debs requires py-enchant

damko:
## Status

Ready for review

## Description of Changes

* Description: 
`make build-debs` requires py-enchant

* Fixes Issue#
I searched through the tickets and I couldn't see any relevant ticket for this

## Testing

no test needed

## Release 

imo this change can go straight to `main` (i.e. latest doc)

## Checklist (Optional)

- [ X] Doc linting (`make docs-lint`) passed locally
- [X] Doc link linting (`make docs-linkcheck`) passed
- [X ] You have previewed (`make docs`) docs at http://localhost:8000

### DIFF
--- a/docs/development/virtual_environments.rst
+++ b/docs/development/virtual_environments.rst
@@ -40,6 +40,7 @@ Debian packages on the staging machines:
 
 .. code:: sh
 
+   sudo apt install python-enchant
    make build-debs
    make staging
    molecule login -s libvirt-staging-focal -h app-staging


conorsch:
Hello, @damko! Do you recall the error you encountered that motivated this change? 

My understanding is that pyenchant is not actually required to build the deb packages. It should be present inside the containers used to build, but that setup happens automatically, and the host machine running the containers needs no knowledge of it. 

To confirm, I made sure both `python-enchant` and `python3-enchant` were absent on my system, and furthermore ran `pip uninstall pyenchant` to remove the package from my local virtualenv (by default present due to requirement at https://github.com/freedomofpress/securedrop/blob/41e79e7214b7aa9416354a64c364e99286423b97/securedrop/requirements/python3/develop-requirements.txt#L530), and `make build-debs` still passed without error. 

If I'm misunderstanding, please let me know, and I'll re-open to discuss further with you!

damko:
> Hello, @damko! Do you recall the error you encountered that motivated this change?

Hi @conorsch! hmm I didn't do anything special, just followed the guide and as soon as I ran `make build-deps` it exited with an error message regarding py-enchant. I installed py-enchant, ran make again and it went ahead. I didn't follow through the whole process mentioned in the testing section because I decided to use production instead. 

> If I'm misunderstanding, please let me know, and I'll re-open to discuss further with you!

no, you are understand it right but I clearly remember that I had to install the package. Maybe it's something due to my own configuration? Btw, I'm running Debian Testing although I don't see how this could influence the need for a python module in a "make" scenario. 
Anyway there is only one way to figure this out: I'll remove py-enchant, repeat the procedure, look into things and come back to you. I just need a bit of time

PR #224, opened 2021-05-18
Update Admin USB rebuild instructions for v3 + 20.04 instances

### LINKED ISSUE #138: Instructions for rebuilding Admin Workstation are not applicable for v3
We have awesome instructions for rebuilding an Admin Workstation here:
https://docs.securedrop.org/en/stable/rebuild_admin.html

Unfortunately, the intro currently notes:
> These instructions refer to a SecureDrop instance using v2 onion services. If your instance uses v3 onion services and you need to rebuild your Admin Workstation, please contact FPF through the SecureDrop Support Portal.

We should update these instructions at the earliest opportunity, as these procedures may be required during the Ubuntu 20.04 migration.

rocodes:
## Status

Ready for review 

## Description of Changes

* Description: Update Admin rebuild instructions for use with Focal / v3 onions

* Fixes #138

## Testing
- visual review
- step through instructions with fresh Admin USB on hardware 
 
## Release 
* n/a 

## Checklist (Optional)

- [:x: ] Doc link linting (`make docs-linkcheck`) passed - error on unrelated link (https://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html), appears to be ssl-related

### DIFF
--- a/docs/rebuild_admin.rst
+++ b/docs/rebuild_admin.rst
@@ -22,8 +22,9 @@ may be simpler. An outline of the steps involved in rebuilding an
  #. Set up SSH access for the new *Admin Workstation*.
  #. Retrieve SecureDrop configuration settings from the *Application* and *Monitor Server*.
  #. Back up and configure the SecureDrop application.
- #. Run the ``./securedrop-admin install`` and ``./securedrop-admin tailsconfig``
-    commands from the new *Admin Workstation*.
+ #. Run ``./securedrop-admin install`` and ``./securedrop-admin tailsconfig``
+    from the new *Admin Workstation*.
+ #. Configure SSH-over-TOR. 
  #. Complete post-rebuild tasks.
 
 
@@ -355,7 +356,7 @@ previous steps. To do so, connect to the Tor network on the
 
 The ``sdconfig`` command will prompt you to fill in configuration details
 about your instance. Use the information retrieved in the previous steps.
-When prompted whether or not to enable SSH-over-Tor, type **yes** (recommended).
+When prompted whether or not to enable SSH-over-Tor, type **no**.
 
 Next, back up the Application server by running the following command in the terminal:
 
@@ -382,10 +383,6 @@ Once the command completes successfully, run
 
 Once this command is complete:
 
- - verify that the Hostname references in ``~/.ssh/config`` have been updated
-   to refer to Onion URLs instead of direct IP addresses
- - verify that you can connect to
-   the servers using ``ssh app`` and ``ssh mon``
  - verify that the desktop shortcuts for the *Source* and *Journalist Interfaces*
    work correctly, opening their respective homepages in Tor Browser.
 
@@ -396,6 +393,39 @@ should reboot the servers, by issuing the following commands in a terminal:
 
  ssh app sudo reboot
  ssh mon sudo reboot
+
+Step 7: Set up SSH-over Tor
+===========================
+
+Rerun the command:
+
+.. code:: sh
+
+ ./securedrop-admin sdconfig
+
+Press "Enter" to use the pre-populated values, but when asked whether to 
+configure SSH-over-Tor, type **yes** (recommended).
+
+Then, re-run
+
+.. code:: sh
+
+ ./securedrop-admin install
+
+When the installation completes, run:
+
+.. code:: sh
+
+ ./securedrop-admin tailsconfig
+
+Once this command completes:
+
+ - verify that the Hostname references in ``~/.ssh/config`` have been updated
+   to refer to Onion URLs instead of direct IP addresses
+ - verify that you can connect to
+   the servers using ``ssh app`` and ``ssh mon``
+ - verify that the desktop shortcuts for the *Source* and *Journalist Interfaces*
+   work correctly, opening their respective homepages in Tor Browser. 
 
 Step 8: Post-rebuild tasks
 ==========================
--- a/docs/rebuild_admin.rst
+++ b/docs/rebuild_admin.rst
@@ -24,7 +24,7 @@ may be simpler. An outline of the steps involved in rebuilding an
  #. Back up and configure the SecureDrop application.
  #. Run ``./securedrop-admin install`` and ``./securedrop-admin tailsconfig``
     from the new *Admin Workstation*.
- #. Configure SSH-over-TOR. 
+ #. Configure SSH-over-TOR.
  #. Complete post-rebuild tasks.
 
 
@@ -397,13 +397,19 @@ should reboot the servers, by issuing the following commands in a terminal:
 Step 7: Set up SSH-over Tor
 ===========================
 
+.. note::
+
+   Without performing this step, you will not be able to access your SecureDrop
+   servers from outside the local network. See  :doc:`ssh_over_local_net`
+   for more information.
+
 Rerun the command:
 
 .. code:: sh
 
  ./securedrop-admin sdconfig
 
-Press "Enter" to use the pre-populated values, but when asked whether to 
+Press "Enter" to use the pre-populated values, but when asked whether to
 configure SSH-over-Tor, type **yes** (recommended).
 
 Then, re-run
@@ -425,7 +431,7 @@ Once this command completes:
  - verify that you can connect to
    the servers using ``ssh app`` and ``ssh mon``
  - verify that the desktop shortcuts for the *Source* and *Journalist Interfaces*
-   work correctly, opening their respective homepages in Tor Browser. 
+   work correctly, opening their respective homepages in Tor Browser.
 
 Step 8: Post-rebuild tasks
 ==========================
@@ -474,7 +480,7 @@ We recommend completing the following tasks after the rebuild:
     ~/Persistent/securedrop/install_files/ansible-base/app-journalist.auth_private
     ~/Persistent/securedrop/install_files/ansible-base/tor-v3-keys.json # for Admin Workstations only
 
-   You may copy these files using a *Transfer Device* (which must be wiped afterwards), 
+   You may copy these files using a *Transfer Device* (which must be wiped afterwards),
    or boot into each of your additional Tails workstations, plug in and unlock your
    *Admin Workstation*'s encrypted partition via the **Places** app, and manually copy
    the file(s) from the Admin Workstation to the same directory on the target Tails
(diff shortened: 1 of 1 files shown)


rocodes:
Comments for reviewer:
- Currently, the backup happens before SSH-over-Tor has been set up again. So the steps are: enable ssh-over-LAN, grab required files, sdconfig, backup, enable ssh-over-Tor, install, tailsconfig.  This seems less than ideal, but I guess it's safer than proceeding with the `install` step without having an existing backup.  I'm kind of hoping someone will vote to move the backup til later in the process. 
- I've included grabbing the journalist alerts ossec configuration info in this rebuild; it was not previously included.

eloquence:
One early observation: The instructions for single-user mode may need to be updated. At least on my NUC10s with Ubuntu 20.04:
- The GRUB menu does not appear unless I hold "Escape" during boot
- The Linux boot options look different:
IMG_20210520_134614

@rocodes would you like to consider that in the scope of this PR, or would you prefer if I file a separate issue for it?

docs/rebuild_admin.rst
```diff
@@ -127,7 +121,7 @@ chosen, you can check as follows:
  #. Log in to the *Application Server* via the console using the adminstration username
     and password.
  #. Check to see if an SSH hidden proxy service exists, using the command
-    ``sudo cat /var/lib/tor/services/ssh/hostname``. If this file exists and
+    ``sudo cat /var/lib/tor/services/sshv3/hostname``. If this file exists and
     includes an Onion URL and authorization token, your instance is set up
```
eloquence, line 125: For v3 onion services, there is no token in this file, it's just the onion address (the authorized public key is in `authorized_clients`).

docs/rebuild_admin.rst
```diff
@@ -195,7 +189,7 @@ First, log on to the *Application Server* via the console and edit the file
 
 Then, restart ``sshd`` using the command ``sudo service sshd restart``.
 
-Finally, check the file ``/etc/network/interfaces``, and note down the network
+Finally, check the file ``/etc/netplan/10-ifupdown.yaml``, and note down the network
```
eloquence, line 192: That file does not exist on my 20.04 app server. I have an `/etc/netplan` directory, and it contains `00-installer-config.yaml`, but no `10-ifupdown.yaml`.

rocodes: Right. `10-ifupdown.yaml` is for those guilty of `netplan migrate`, but not clean install, and that's the last state my servers were in. Shall update.

rocodes:
Thanks for your review! I've addressed the two inline comments but not the grub/single-user-mode issue in the NUC10s in 82d2fe2, and the GRUB issues in 8c913bb (I did update instructions about the kernel opts, but didn't provide updated instructions for getting into the GRUB menu because it does vary based on hardware model and I think it's fair to expect admins to figure that part out themselves if need be).

rocodes:
Thanks for your review! I've addressed your comments in a5a6392. 

I didn't provide updated instructions for getting into the GRUB menu because it does vary based on hardware model and I think it's fair to expect admins to figure that part out themselves if need be, but I did update the line about the kernelopts in single user mode.

eloquence:
I stepped through a full rebuild using these docs and while I encountered a couple of stumbles (I specified the wrong IP to `iptables`; later Ansible errored out once during "Timed out waiting for last boot time check (timeout=600)"; I had to restart Tor manually to regain SSH access), none of this was due to the instructions here, which worked beautifully. :tada:   

I may push some minor tweaks but otherwise I think this looks pretty good to go.

docs/rebuild_admin.rst
```diff
@@ -172,8 +167,9 @@ and deleting the line:
 
 Then, restart ``sshd`` using the command ``sudo service sshd restart``.
 
-Finally, check the file ``/etc/network/interfaces``, and note down the network
-settings for the default Ethernet interface. You'll need them in the next step.
+Finally, check the file ``/etc/netplan/00-installer-config.yaml``, and note
```
eloquence, line 170: As far as I can tell, all we need in the next step is the IP address of both servers. Is there a reason not to use the simpler `ip addr` (`ip a`) command here, particularly considering that the location of this file might change in the course of administration?

docs/rebuild_admin.rst
```diff
@@ -379,54 +368,65 @@ previous steps. To do so, connect to the Tor network on the
...
 
- ./securedrop-admin backup
+ ./securedrop-admin sdconfig
 
+Press "Enter" to accept the prepopulated values until you reach the
+SSH-over-Tor settings, this time typing **yes** to enable SSH-over-Tor.
 
-Step 6: Run the ``./securedrop-admin install`` command
-======================================================
+Step 7: Use the installer to complete the configuration
```
eloquence, line 397: I changed this section title from "fetch the configuration" in 90668bbda80ecd95f4f222d31767c01c6b7b2242 because the installation playbook may also change the Tor onion service authentication details present on the server, making it more than a fetch operation.

eloquence:
This looks good from my end! @zenmonkeykstop, it would be great if you could have a look. I think a visual review & input on overall structure and open comments is probably sufficient; I tested the overall process (except for the HTTPS configuration -- I only tested that the `tar` syntax works), and it worked well for me.

docs/rebuild_admin.rst
```diff
@@ -242,85 +238,78 @@ In addition to the account and networking information retrieved from the servers
...
-
-If your instance was set up to use SSH over TOR, you'll also need to copy over the details
-of the SSH proxy onion services, by running the following commands:
-
-.. code:: sh
-
- echo "HidServAuth $(ssh app sudo cat /var/lib/tor/services/ssh/hostname)" > app-ssh-aths
- echo "HidServAuth $(ssh mon sudo cat /var/lib/tor/services/ssh/hostname)" > mon-ssh-aths
+ echo "$(ssh app sudo cat /var/lib/tor/services/journalistv3/hostname)" \
+    > app-journalistv3-ths
```
zenmonkeykstop, line 256: This file isn't used (see comment above). The server doesn't have the private key necessary to build the `app-*.auth_private` files needed by `tailsconfig` to set up access.

zenmonkeykstop (dismissed):
I haven't attempted a run-thru yet, but I think this needs some changes to address the fact that authentication keys will change. At minimum all existing keys need to be reprovisioned (this wasn't the case for v2 because the hidden service auth token didn't change).

docs/rebuild_admin.rst
```diff
@@ -379,54 +368,65 @@ previous steps. To do so, connect to the Tor network on the
...
+To revert the changes made to enable temporary local SSH access, you
 should reboot the servers, by issuing the following commands in a terminal:
 
 .. code:: sh
 
  ssh app sudo reboot
  ssh mon sudo reboot
 
-Step 7: Post-rebuild tasks
+Step 8: Post-rebuild tasks
```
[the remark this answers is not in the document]
eloquence: Can confirm, had to update my other workstations after running through the rebuild process.

rocodes:
(update: more work tk on this on Monday)

docs/rebuild_admin.rst
```diff
@@ -379,54 +368,65 @@ previous steps. To do so, connect to the Tor network on the
 .. note:: The ``./securedrop-admin setup`` command may take several minutes to complete, and may
  fail due to network issues. If it fails, it's safe to run again.
 
-When prompted, fill in the values retrieved in the previous steps.
+The ``sdconfig`` command will prompt you to fill in configuration details
+about your instance. Use the information retrieved in the previous steps.
+When prompted whether or not to enable SSH-over-Tor, type **no** to ensure
```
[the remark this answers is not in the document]
rocodes: Thank you, yes- fixed in 33f6c42

docs/rebuild_admin.rst
```diff
@@ -242,85 +238,78 @@ In addition to the account and networking information retrieved from the servers
...
-If your instance was set up to use SSH over TOR, you'll also need to copy over the details
-of the SSH proxy onion services, by running the following commands:
-
-.. code:: sh
-
- echo "HidServAuth $(ssh app sudo cat /var/lib/tor/services/ssh/hostname)" > app-ssh-aths
- echo "HidServAuth $(ssh mon sudo cat /var/lib/tor/services/ssh/hostname)" > mon-ssh-aths
+ echo "$(ssh app sudo cat /var/lib/tor/services/journalistv3/hostname)" \
+    > app-journalistv3-ths
+ echo "$(ssh app sudo cat /var/lib/tor/services/sourcev3/hostname)" > app-sourcev3-ths
```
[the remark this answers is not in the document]
rocodes: Have moved it to `/tmp/sourcev3` in 33f6c42, and only grab the SI onion address.

docs/rebuild_admin.rst
rocodes: Right, thanks, this is unnecessary, and I've removed it in 33f6c42

docs/rebuild_admin.rst
```diff
@@ -242,85 +238,78 @@ In addition to the account and networking information retrieved from the servers
...
- - GPG *Submission Public Key* and OSSEC public key
+ - GPG *Submission Public Key*, *OSSEC Alert Public Key*, and (optional)
+   *Journalist Alert Public Key*
  - OSSEC alert configuration details
  - (Optional) HTTPS configuration details
 
-Retrieve Onion Service info
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
+Retrieve Onion Service Info
```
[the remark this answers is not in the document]
rocodes: Thanks for clarifying. This makes sense and you're right, running through `install` then `tailsconfig` does regenerate all the right files and put them in place. I've removed this section.

docs/rebuild_admin.rst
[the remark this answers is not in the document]
rocodes: Thanks for the reminder - addressed in 33f6c42

eloquence:
(Starting another round of review on this today, may not get all the way through until tomorrow.)

eloquence:
I stepped through the whole process one more time and I think this is good to go.

I hit one snag: When running `./securedrop-admin install`, it kept rebooting the servers, which kept destroying the `iptables` rules, causing subsequent commands to fail. I'm not quite sure why it kept rebooting, after the task "Add overrides for apt-daily timers", with Ansible then including the reboot.yml file. Commenting that file out let me proceed, but I'd like to understand if this is an error real users may encounter, and if so, how we may be able to mitigate it.

eloquence (approved):
Thanks @rocodes! As per our discussion last week, I've not re-tested whether the SSH-over-LAN setting resolves the reboot issue during the playbook run, but logically it makes sense to me, and I trust in your own test results. @zenmonkeykstop, I believe your previous review comments were all addressed - I think this is good to merge. 

Note I've appended one small commit to clarify the implications of the SSH-over-Tor vs. SSH-over-LAN setting.

Thanks a lot for the hard work on this, glad to be getting these updated docs in.

PR #225, opened 2021-05-19
Remove references to the flag-for-reply workflow.

rmol:
## Status

Ready for review

## Description of Changes

Updates the docs after removal of the flag-for-reply workflow in https://github.com/freedomofpress/securedrop/pull/5954.

## Testing

Not much to check with deletions, but I am interested in whether others think we should retain the API endpoint documentation. I decided it was just confusing, since the endpoint now does nothing.

## Release 

This should be merged after https://github.com/freedomofpress/securedrop/pull/5954.

## Checklist (Optional)

### DIFF
--- a/docs/development/journalist_api.rst
+++ b/docs/development/journalist_api.rst
@@ -553,23 +553,6 @@ Response 200:
     "message": "Star removed"
   }
 
-Flag a source
-^^^^^^^^^^^^^
-
-Requires authentication.
-
-.. code:: sh
-
-  POST /api/v1/sources/<source_uuid>/flag
-
-Response 200:
-
-.. code:: json
-
-  {
-    "message": "Source flagged for reply"
-  }
-
 Submissions
 -----------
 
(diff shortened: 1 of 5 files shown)


eloquence:
LGTM in principle once the main PR lands.

> I am interested in whether others think we should retain the API endpoint documentation. I decided it was just confusing, since the endpoint now does nothing.

As a matter of good practice, it may be worth noting in its own section, such as "Deprecated functionality", explaining that the no-op endpoint and the `is_flagged` property will be removed in a future release. Right now, the presence of the `is_flagged` property in the documentation is a bit confusing, given that it is always `false`.

rocodes (commented):
LGTM. The only thing I would maybe suggest is that there are a couple sample JSON responses that include `is_flagged: false` and it could be good to add a comment (even inline in the code block) explaining that this is a legacy field that can be ignored.  

e.g. https://github.com/freedomofpress/securedrop-docs/pull/225/files#diff-466879439de016d374b3dc3cc22f5e587900a70cfdd5c80a1e51e35227510419L197-R197

eloquence:
I've added a section to the end for now - @rocodes ready for another look from you

rocodes (approved):
WFM. Thank you @eloquence

PR #229, opened 2021-05-29
Fix subnet recommendation for staging in Qubes OS

### LINKED ISSUE #209: The network setup instructions for staging on Qubes OS don't match the Ubuntu installer
## Describe the change

Update the network setup instructions in the docs that describe how to set up a **staging environment** using **Qubes OS**, so that:

- the fields described in the [Qubes OS docs][docs-qubes] match what he Ubuntu (Focal) installer displays 
- the way the network settings are describes in the Qubes OS docs matches the [non-Qubes OS docs][docs] (which already match Focal's installer display themselves since https://github.com/freedomofpress/securedrop-docs/commit/18ec6f68e398ee0a25d63955b5284d153bcc6024)

_I highlighted "Gateway" in the links so that the relevant sections are easier to find._

The expected fields are:

- Subnet
- Address
- Gateway
- Name servers
- Search domains

  [docs-qubes]: https://docs.securedrop.org/en/stable/development/qubes_staging.html?highlight=Gateway
  [docs]: https://docs.securedrop.org/en/stable/servers.html?highlight=Gateway

When setting up a staging environment on a Qubes OS machine, the instructions to set up Ubuntu will be accurate / easier to follow.

## User Research Evidence

Anecdotical evidence of myself going through the docs and having to stop and think about what the Ubuntu installer was asking me.

### LINKED ISSUE #211: The network setup instructions for staging on Qubes OS can be miselading
## Context

In order to setup the network of a fresh Ubuntu (Focal) installation, a **subnet**, **address** and **gateway** must be provided.

The values for those three fields are interdependent, and the assumptions built into the current instructions are sometimes not fulfilled.

Specifically:

- the IP of the `sd-staging-base-focal` VM is assigned when the VM is created. Because of the the docs recommend reading its value via `qvm-prefs sd-staging-base-focal ip`. (Which is not wrong a all in itself.)
- the IP of the gateway that is visible to `sd-staging-base-focal` can also vary. Because of the the docs recommend reading its value via `qvm-prefs sd-staging-base-focal visible_gateway`. (Which, again, is not wrong a all in itself.)
- the suggested value for the netmask is suffix is `/24`. **This carries assumptions on the two values above.**

As far as I can tell, my Qubes OS network settings are the default (I use R4.1), despite that, *the assumption that the gateway would belong to the same `/24` network as the machine was not upheld in my case.*

- Address: `10.137.0.34` (`sd-staging-base-focal`)
- Gateway: `10.138.13.65` *which doesn't belong to the `10.137.0.0/24` subnet*
- Subnet: `10.136.0.0/14` is the smallest subnet that contains both `sd-staging-base-focal` and its gateway. ~(It can be written as `10.137.0.0/14` if that makes things easier to explain.)~

## Describe the change

I'm not sure what's the best balance between giving straightforward instructions and giving optimal instructions. Calculating the _subnet_ based on the _address_ and _gateway_ would give a correct and optimal solution, but adds steps to the setup.

I am currently [asking about reference Qubes OS documentation][ask] ([obsolete link][ask-old]) on the IP ranges that are used by default.
If suggesting a predefined suffix in the SecureDrop docs, I think it would be good to choose one that is broad enough to prevent as much network troubleshooting as possible given a fresh Qubes OS installation.

**Edit**: I think that recommending `10.136.0.0/14` would be reasonable given how Qubes OS allocates IP addresses to VMs (see comment below).

  [ask-old]: https://qubes-os.discourse.group/t/3870
  [ask]: https://forum.qubes-os.org/t/default-ip-ranges-for-sys-firewall-appvms-etc/3870

After following the current instructions, my `sd-staging-base-focal` machine couldn't reach the internet to fetch updates. The intent behind providing specific instructions in the docs is to avoid troubleshooting when setting up servers.

## User Research Evidence

Anecdotical evidence based on my own experience setting up staging servers in Qubes OS.

## Additional context

Related to #209

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

- Updates the list of fields that are required by the Ubuntu 20.04 (Focal) installer
- Rename the _hostname_ field as named in the installer, and move it down because it is required in a later step to the network settings
- Modifies the subnet recommendation in for staging in Qubes OS in order to ensure that the resulting machines have access to the network

Fixes #209 
Fixes #211 

## Testing

- [ ] Review the list of fields required by the installer
- [ ] Check that `10.137.0.0/24` indeed contains the likely IPs of both `sd-staging-base-focal` and its gateway

## Release 

_No special considerations for release come to my mind._

## Checklist (Optional)

### DIFF
--- a/docs/development/qubes_staging.rst
+++ b/docs/development/qubes_staging.rst
@@ -65,12 +65,12 @@ Next, choose **Install Ubuntu**.
 For the most part, the install process matches the
 :ref:`hardware install flow <install_ubuntu>`, with a few exceptions:
 
-  -  Server IP address: use value returned by ``qvm-prefs sd-staging-base-focal ip``,
-     with ``/24`` netmask suffix
-  -  Gateway: use value returned by ``qvm-prefs sd-staging-base-focal visible_gateway``
-  -  For DNS, use Qubes's DNS servers: ``10.139.1.1`` and ``10.139.1.2``.
-  -  Hostname: ``sd-staging-base-focal``
-  -  Domain name should be left blank
+  -  **Subnet:** 10.137.0.0/24
+  -  **Address:** use value returned by ``qvm-prefs sd-staging-base-focal ip``
+  -  **Gateway:** use value returned by ``qvm-prefs sd-staging-base-focal visible_gateway``
+  -  **Name servers:** 10.139.1.1,10.139.1.2
+  -  **Search domains:** *should be left blank*
+  -  **Your server's name:** ``sd-staging-base-focal``
 
 Make sure to configure LVM and use **Virtual disk 1 (xvda 20.0GB Xen Virtual Block device)**
 when asked for a target partition during installation. It should be the default option.
--- a/docs/servers.rst
+++ b/docs/servers.rst
@@ -206,7 +206,7 @@ If you did not, adjust these settings accordingly.
   -  **Address:** 10.20.2.2
   -  **Gateway:** 10.20.2.1
   -  **Name servers:** 8.8.8.8, 8.8.4.4
-  -  **Search Domains:** *should be left blank*
+  -  **Search domains:** *should be left blank*
 
 -  *Monitor Server*:
 
@@ -214,7 +214,7 @@ If you did not, adjust these settings accordingly.
   -  **Address:** 10.20.3.2
   -  **Gateway:** 10.20.3.1
   -  **Name servers:** 8.8.8.8, 8.8.4.4
-  -  **Search Domains:** *should be left blank*
+  -  **Search domains:** *should be left blank*
 
 Select **Save** and press **Enter** to apply your settings. Then select **Done** and press **Enter**.
 


gonzalo-bulnes:
@creviera I'm curious about your network configuration in Qubes OS. Maybe we could review this together :slightly_smiling_face:

conorsch:
Hey @gonzalo-bulnes ! This one's been open a whlie, we just dug it up again at sprint-planning earlier today. I offered to take a first pass at it, because I wanted to ask you whether the networking settings were tested on Qubes 4.0 or 4.1. That issude aside, I ended up reading up on some networking settings here and share my thoughts below. 

One of the recommended changes here includes declaring the subnet for the VM are `10.136.0.0/14`, which intuitively strikes me as far too large: 2^18, or 262,144. Turns out I don't know what the proper space is here, so I tried to look it up.

The question I'm trying to answer is "What is the CIDR subnet encompassing IPs for Qubes VMs?"


I've read https://www.qubes-os.org/doc/networking/ and grepped through the code in https://github.com/QubesOS/qubes-core-admin/blob/ee34eb2a936c47c6e20a51d0ac0701929990511d/qubes/vm/mix/net.py#L168-L179 and I don't see any mention of `10.136.*` IPv4 addrs at all! The code does mention `10.137.*` and `10.138.*`, however. Note the bitwise shift operators in the Qubes network code, as well as the commented observation that the AppVM's xen id (qid) is used to determine the IP address. Which is a fine way to ensure we don't have collisions. Translated into a Python one-liner for demonstration purposes:

```
python3 -c 'for i in range(0, 255): print("10.137.{}.{}".format((i >> 8) & 0xff, i & 0xff))'
```

That's the same as:

```
$ netmask -c -r 10.137.0.0/24
     10.137.0.0-10.137.0.255    (256)
```

As an aside, apparently DispVMs use `10.138.0.1/24`, so it's possible to tell whether a VM is Disposable by looking at its internal IP. Neat! But for the purpose of these docs, we don't care about that. The docs recommended a StandaloneVM for simplicity, but I'm actually using an AppVM myself. And since `10.137.0.1/24` and `10.138.0.1/24` are not contiguous, there's no point in trying to unify them in a single subnet declaration. 

Looks like the `10.137.0.1/24` is the correct answer to our question, and that's what we should include in the docs. 

Does the above look accurate to you, @gonzalo-bulnes? If so, do you agree we should tweak that line in this PR? That was the only quibble I had with the content—the rest of the diff looks just swell to my eye.

gonzalo-bulnes:
> The question I'm trying to answer is "What is the CIDR subnet encompassing IPs for Qubes VMs?"
:white_check_mark: That's and was my intention too.

Not including the DispVM. Seems all right to me. I can't think of a DispVM that we'd need to connect to _through the network_ (RPC are more likely for passing documents around, am I correct?).

Wouldn't the AppVMs need to be able to reach the name severs though? Those are assigned `10.139.1.1` and `10.139.1.2`. I can look up where I got that from, but I did confirm it, maybe as part of https://qubes-os.discourse.group/t/3870

gonzalo-bulnes:
Also, answering your initial question @conorsch, I tested this on R4.1. I don't believe there have been changes in that part of Qubes OS between R4.0 and R4.1.

gonzalo-bulnes:
Did a two things at once:
- Updated the branch (rebased)
- Changed the subnet recommendation to `10.137.0.0/24`

@conorsch If you don't feel strongly about `10.137.0.1/24`, I find a little easier on the mind to keep the segments that are variable in a subnet to `0`. That being said I don't have strong feelings about that either.

Also, I'd favor `10.137.0.0/24` over "the value returned by `qvm-prefs sd-staging-base-focal ip`, with `/24` netmask suffix" because we know the gateway will be in the `10.137.0.0/24` range anyway. (The docs are long enough to try and go to the point IMHO.)

gonzalo-bulnes:
Thanks for the thoughtful and thorough review @conorsch!

conorsch (approved):
Love it. Thanks for the in-depth discussion, @gonzalo-bulnes, I learned a lot as part of review here! :rainbow:

PR #232, opened 2021-06-16
Add 1.8.2->2.0.0 update guide; bump version

eloquence:
## Status
Ready for review, in draft state to merge on day of release

## Description of Changes

- Adds standard update guide
- Includes note about Tails updates requiring manual workaround
- Assumes that 2.0.0 tag will be signed using the _old_ key; see https://github.com/freedomofpress/securedrop/issues/5923#issuecomment-862796305
- Adds notice in a few place informing folks of the key transition and linking to our transition statement

## Checklist (Optional)


docs/upgrade/1.8.2_to_2.0.0.rst
```diff
@@ -1,19 +1,18 @@
-Upgrade from 1.7.0 to 1.7.1
+Upgrade from 1.8.2 to 2.0.0
```
eloquence, line 1: This is git's move detection being off and treating a new file as an edit. :/

eloquence: Split into its own commit as fae248d27de83b4432f27e3fcfd33d7a54b72010 to aid diff readability.

rocodes:
Have reviewed (it's still in draft mode), LGTM with one or two small suggestions:
- If a user is on Tails 4.14 or earlier, I think we should just advise them to manually upgrade. I wonder if it's worth calling that case out specifically or not.   
- Maybe we should include an explicit nod to Xenial stragglers in the Updating Servers section? "If your instance is running Ubuntu 16.04, your SI has been disabled and you will not receive this update. Contact us for assistance."  

- (out of scope just mentioning) We should maybe consider updating the "Getting Support" text to clarify that, if you run a SecureDrop instance, you should sign up for an account on the support portal, rather than filing issues on the community forum (as the wording sort of suggests). But I can do that in another PR.

eloquence:
Thanks @rocodes for the review!

> Maybe we should include an explicit nod to Xenial stragglers in the Updating Servers section?

Done.

> If a user is on Tails 4.14 or earlier, I think we should just advise them to manually upgrade. I

I still had a Tails 4.2 drive, so I just tested the workaround in this PR. It triggered an automatic update to 4.19 as expected, so I think as long as folks are at least on the Tails 4 series, we can recommend they run the one-liner.

> We should maybe consider updating the "Getting Support" text 

Agree, a separate PR for that would be awesome - thanks for flagging. :)

rocodes (approved):
LGTM! Thank you @eloquence for preparing.

Issue #234, opened 2021-06-24
Images failing to load in live builds


eloquence:
Appears LFS-related, see journalist guide for example:

https://docs.securedrop.org/en/stable/journalist.html

Example image URL: https://docs.securedrop.org/en/stable/_images/journalist-delete_submissions.png

The actual file contents are an LFS pointer, so something appears to be going wrong with the LFS setup in `conf.py`

zenmonkeykstop:
Latest failing build log is here: https://readthedocs.org/api/v2/build/14089116.txt

gonzalo-bulnes:
It seems to me that this part of the log is relevant:

```log
2021-06-24 17:24:42 (19.3 MB/s) - ‘git-lfs-linux-amd64-v2.12.0.tar.gz’ saved [4675677/4675677]

README.md
CHANGELOG.md
git-lfs
install.sh
Updated git hooks.
Git LFS initialized.
fetch: Fetching reference HEAD
Error updating the git index:
git-lfs filter-process: 1: git-lfs filter-process: git-lfs: not found
error: packet write with format failed
error: Could not write client identification
error: initialization for subprocess 'git-lfs filter-process' failed
fatal: docs/diagrams/SecureDrop-0.3-DFD.png: clean filter 'lfs' failed

Checking out LFS objects: 100% (228/228), 20 MB | 0 B/s, done.

Errors logged to /home/docs/checkouts/readthedocs.org/user_builds/securedrop-docs/checkouts/latest/.git/lfs/logs/20210624T172446.320040923.log
Use `git lfs logs last` to view the log.
```

eloquence:
@gonzalo-bulnes Pretty sure it's the LFS setup failing; note how images like https://docs.securedrop.org/en/stable/_images/journalist-delete_source_account.png resolve to LFS pointers:

```
$ curl https://docs.securedrop.org/en/stable/_images/journalist-delete_source_account.png

version https://git-lfs.github.com/spec/v1
oid sha256:b8dd91d867a9886050dd986516d3eeff8131697623cb92551132fbbc2b7ab119
size 127752
```
This from the build log

> git-lfs filter-process: 1: git-lfs filter-process: git-lfs: not found

suggests to me that the installation of `git-lfs` is failing. Note that RTD does not support LFS out of the box; this is bolted on in our `conf.py` here:

[...]

PR #237, opened 2021-06-27
Mitigate absence of images in the User Guides

### LINKED ISSUE #236: Parts of the documentation are fragile with respect to missing images
## Current situation

Currently, some images are essential to the documentation. That happens in several ways:

1. Images that convey information that is not present in the surrounding text
2. Text that makes explicit reference to images (which are not always available):
  
A good example of both cases is:
  > After you delete the message from the journalist, make sure you see the
below message.
  > Delete received messages

3. `alt` text that doesn't integrate well with the surrounding copy (when images are not displayed for, the `alt` text in rendered in their place). That causes the text to make _less_ sense than it would if the image / `alt` text was entirely missing.

> Click Save file. In the save dialog, select one of the two folders highlighted in red in the screenshot below:
> Download to sandbox folder
> The difference between these two folders [...]

## Problem

When essential images are not displayed ([screen reader](https://github.com/freedomofpress/securedrop-docs/issues/70), [outage](https://github.com/freedomofpress/securedrop-docs/issues/234)) the documentation becomes significantly less useful.

## Opportunity

Images could be used to _enhance_ the documentation, while ensuring that the combination of surrounding text and `alt` text provides all the essential information.

One way to think about it is to treat images as an _alternative_ way to consume the content. If a confirmation message is essential, its content is likely relevant. Readers could be instructed to look for "Message sent." on a given page, while provided with a screenshot that displays such message in the context of that page.

**Note**: Whether they can see them or not, people using screen readers are aware of what images are! The presence of a screenshot in a piece of documentation isn't unexpected, and there is no need to _hide_ the presence of images when they're not displayed, or avoid repetition too aggressively. Suffice to ensure that their presence is clear, and the information they provide is available in their absence. In other words, when thinking about `alt` text, the background of a button is an _illustrative_ image for which an empty value is adequate, while a screenshot is likely not (even though the information it conveys may be available in the surrounding text).

The documentation would become more accessible to everyone:

- it would be accessible more often, even when outages prevent images from being rendered
- it would be accessible from more devices and more situations (e.g. Narrator, TalkBack, VoiceOver, Orca, sunny outdoor environments...)

## User Research Evidence

As an approach, the proposed remediation / described opportunity is an extension on the well-documented [progressive enhancement](https://en.wikipedia.org/Progressive_enhancement) strategy for web design. That perspective helps spotting instances of (2).

I personally noticed myself instances of (1) while setting up firewall rules a couple of days ago. Even though my reading device had the ability to display images, they were missing due to an unexpected outage and some configuration information became inaccessible.

I've also experienced (3) in different occasions / websites when using VoiceOver.

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

Implements the approach proposed in #236 for the sections of the documentation that target end-users (sources, journalists, admins).

Instead of coming up with new ways of composing the text with the images, I've re-applied as much as possible the patterns that I've found achieved the purpose in other parts of the documentation.

~~Fixes~~ Contributes to fixing #236
Mitigates the impact of #234 
Contributes to fixing #70 

_Compatible with (but does not depend on, nor require) https://github.com/freedomofpress/securedrop-docs/pull/196_

## Testing

- [ ] Proof-read the paragraphs around each modification to ensure continuous reading of text and images makes sense.
- [ ] Do the same without displaying the images (e.g. delete the images locally).

## Release 

 
_No special considerations come to mind for release._ 

## Checklist (Optional)

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -394,7 +394,7 @@ only want to receive text messages, you can disable uploads as follows:
 #. Click **Update Submission Preferences**
 
 This change will be applied immediately on the *Source Interface*. Documents that
-were previously uploaded will still be available via the **Journalist Interface**.
+were previously uploaded will still be available via the *Journalist Interface*.
 
 .. _test OSSEC alert:
 
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -250,20 +250,20 @@ smartphone authenticator app or a Yubikey for two-factor authentication. If an
 app is to be used, the journalist should install it before proceeding with the
 account setup.
 
-  .. include:: includes/otp-app.txt
+.. include:: includes/otp-app.txt
 
-* Click **Admin** in the top right corner of the page to load the *Admin Interface*.
+#. Click **Admin** in the top right corner of the page to load the *Admin Interface*.
 
-  |SecureDrop admin home|
+   |SecureDrop admin home|
 
-* Click **Add User** to add a new user.
+#. Click **Add User** to add a new user.
 
-  |Add a new user|
+   |Add a new user|
 
-* Hand the keyboard over to the journalist so they can create their own username.
-* Once they’re done entering a username for themselves, have them save their pre-generated diceware passphrase to their password manager.
-* If the new account should also have admin privileges, allowing them to add or delete other journalist accounts, select **Is Admin**.
-* Finally, set up two-factor authentication for the account, following one of the two procedures below for your chosen method.
+#. Hand the keyboard over to the journalist so they can create their own username.
+#. Once they’re done entering a username for themselves, have them save their pre-generated Diceware passphrase to their password manager.
+#. If the new account should also have admin privileges, allowing them to add or delete other journalist accounts, select **Is Admin**.
+#. Finally, set up two-factor authentication for the account, following one of the two procedures below for your chosen method.
 
 .. note::
    The username **deleted** is reserved, as it is used to mark accounts which
@@ -273,13 +273,13 @@ account setup.
 FreeOTP
 #######
 
-* If the journalist is using FreeOTP or another app for two-factor authentication, click **Add User** to proceed to the next page.
+#. If the journalist is using FreeOTP or another app for two-factor authentication, click **Add User** to proceed to the next page.
 
    |Enable FreeOTP|
 
-* Next, the journalist should open FreeOTP on their smartphone and scan the barcode displayed on the screen.
-* If they have difficulty scanning the barcode, they can tap on the icon at the top that shows a plus and the symbol of a key and use their phone's keyboard to input the two-factor secret into the ``Secret`` input field, without whitespace.
-* Inside the FreeOTP app, a new entry for this account will appear on the main screen, with a six-digit number that recycles to a new number every thirty seconds.  The journalist should enter the six-digit number in the  **Verification code** field at the bottom of the **Enable FreeOTP** form and click **Submit**.
+#. Next, the journalist should open FreeOTP on their smartphone and scan the barcode displayed on the screen.
+#. If they have difficulty scanning the barcode, they can tap on the icon at the top that shows a plus and the symbol of a key and use their phone's keyboard to input the two-factor secret into the ``Secret`` input field, without whitespace.
+#. Inside the FreeOTP app, a new entry for this account will appear on the main screen, with a six-digit number that recycles to a new number every thirty seconds.  The journalist should enter the six-digit number in the  **Verification code** field at the bottom of the **Enable FreeOTP** form and click **Submit**.
 
 If two-factor authentication was set up successfully, you will be redirected back
 to the *Admin Interface* and will see a confirmation that the two-factor code was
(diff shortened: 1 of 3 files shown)


gonzalo-bulnes:
Updating the **Install SecureDrop** is less straightforward, so I'll limit this PR to the **User Guides** to make review easier.

docs/admin.rst
```diff
@@ -250,52 +250,36 @@ smartphone authenticator app or a Yubikey for two-factor authentication. If an
...
-  .. include:: includes/otp-app.txt
+.. include:: includes/otp-app.txt
 
-* First, click **Admin** in the top right corner of the page to load the *Admin Interface*:
+#. Click **Admin** in the top right corner of the page to load the *Admin Interface*.
 
   |SecureDrop admin home|
 
-*  Once there, click **Add User** to add a new user:
+#. Click **Add User** to add a new user.
```
eloquence, line 259: The way it's currently formatted, this will restart the numbering at 1.

docs/admin.rst
```diff
@@ -250,52 +250,36 @@ smartphone authenticator app or a Yubikey for two-factor authentication. If an
...
 
-|Add a new user|
+  |Add a new user|
 
-* Next, hand the keyboard over to the journalist so they can
-  create their own username.
+#. Hand the keyboard over to the journalist so they can create their own username.
```
eloquence, line 263: Numbering is restarting again here.

docs/journalist.rst
```diff
@@ -150,33 +151,32 @@ Step 1: Download the encrypted submission
...
 
-A dialog box with two choices will appear, **Cancel** and **Save file**:
+A dialog box with two choices will appear, **Cancel** and **Save file**.
 
 |Download selected|
 
-Click **Save file**. In the save dialog, select one
-of the two folders highlighted in red in the screenshot below:
+Click **Save file**. In the save dialog, select one of the two folders
+called **Tor Browser** and **Tor Browser (persitent)**.
```
eloquence, line 167: typo: persitent->persistent

docs/journalist.rst
```diff
@@ -594,7 +594,7 @@ Decrypting and Preparing to Publish
 .. note::
 
    To decrypt a VeraCrypt drive on a Windows or Mac workstation, you need
-   to have the VeraCrypt software installed. If you are unsure if you have the
+   to have the **VeraCrypt** software installed. If you are unsure if you have the
```
eloquence, line 597: The bolding here seems a bit excessive, I think unless we are explicitly directing folks to a menu or part of the UI as part of referencing the name of the application, we can omit it (applicable to other additions of bolding for software names as well)

docs/source.rst
```diff
@@ -265,28 +268,40 @@ correspondences you had with journalists.
...
-below message.
+After you delete the reply from the journalist, make sure you see the
+confirmation message: "Reply deleted".
 
 |Delete received messages|
 
 .. |Source Interface Security Slider Warning| image:: images/manual/securedrop-security-slider-warning.png
+   :alt: Warning banner: Your Tor Browser's Security Level is too low.
 .. |Security Slider| image:: images/manual/source-turn-slider-to-high.png
+   :alt: Advanced Security Settings in Tor Borwser.
```
eloquence, line 279: typo: Borwser->Browser

docs/admin.rst
```diff
@@ -823,12 +800,20 @@ left off. However, if the same issue persists, you will need to investigate
 further.
 
 .. |Reset Passphrase| image:: images/manual/screenshots/journalist-edit_account_user.png
+   :alt: The account edition form allows change name, reset passphrase, and reset two-factor authentication.
```
eloquence, line 803: - edition form->editing form
- allows change name->allows admins to change name

eloquence (changes requested):
Thanks a lot for your work on this, @gonzalo-bulnes. I really like the general accessibility principle you are applying here of making sure that the text speaks for itself, without relying on images the user may not be able to see.

The addition of descriptive ALT text is a huge improvement, as well, and switching to numbered lists instead of bullet lists for step-by-step instructions makes a lot of sense.

I noted a few small issues, and will need to take another careful read through since it's a fairly large diff, but I'm optimistic we can land this next week if you have time to work on it. 

I suggest you rebase as well, to pick up the removal of LFS. I rebased locally and it seemed to apply without issue.

docs/journalist.rst
gonzalo-bulnes: That makes sense, I'll review the other instances too.

docs/admin.rst
gonzalo-bulnes: Good catch. I'll fix that, I think the blank lines may be at fault but I'll figure that out.

gonzalo-bulnes:
Thank you for you review @eloquence! I clearly missed a spellcheck run, sorry about that. I've got time tomorrow to address your comments and rebase the branch, I'll do that :slightly_smiling_face:

docs/journalist.rst
gonzalo-bulnes: Done.

docs/source.rst
gonzalo-bulnes: Fixed.

docs/admin.rst
gonzalo-bulnes: Done.

docs/journalist.rst
gonzalo-bulnes: Fixed.

docs/admin.rst
gonzalo-bulnes: Fixed. It wasn't the blank lines, but the indentation of the image references. Four spaces (not 3) are needed for the image to be considered part of the list item. If the images aren't considered as part of the list items, a new list starts after each image.

docs/admin.rst
gonzalo-bulnes: Fixed.

gonzalo-bulnes:
I applied the suggestions, performed a general spelling check and rebased.

eloquence (approved):
Looks great, thanks again. (In case it's helpful for others: I used a [bookmarklet for converting ALT attributes to TITLE attributes](https://www.howtocreate.co.uk/bookmarklets.html) to preview the ALT attributes as tooltips.)

PR #241, opened 2021-07-07
Add 2.0.1 upgrade guide; bump version

rocodes:
## Status

Ready for review

## Description of Changes

- add upgrade guide for 2.0.1
- bump version

## Testing
* visual review

## Release 
* to be merged when 2.0.1 is released 

## Checklist (Optional)

- [x] Doc linting (`make docs-lint`) passed locally
- [:x:] Doc link linting (`make docs-linkcheck`) passed No, but the link error is from an unrelated portion of docs, I'll do a separate fixup PR
- [x] You have previewed (`make docs`) docs at http://localhost:8000

PR #249, opened 2021-07-22
Remove inaccurate includes about signing key expiry

eloquence:
## Status
Ready for review 

## Description of Changes

#241 did not remove the instructions noting that the old release key expired from some sections where they are no longer applicable, because they refer to the new (not-expired) release key. This PR fixes that.

## Checklist (Optional)


conorsch (approved):
LGTM. Now the only place the include is used is `/upgrade/1.8.2_to_2.0.0.html`, and it's still relevant there.

PR #258, opened 2021-10-07
Fix YubiKey linkcheck issues

eloquence:
- The linkcheck tool hits a 403 Cloudflare page on linkcheck runs,
  but it's accessible just fine via browsers. Seems like
  ignore-listing it is our best bet for now.

- The technical manual also 403s, and it includes a note redirecting
  to a new URL, which does not have that issue.

## Status

Ready for review
## Checklist (Optional)


cfm (approved):
Configuration change looks good.  Documentation change checks out via `make docs`.

PR #274, opened 2021-11-15
Remove inaccurate tip in inauspicious location

eloquence:
## Status
Ready for review 

## Description of Changes

Removes a tip that instructs admins to copy nonexistent (because v2) configuration files. I opted to not update it because this seems like a location that's prone to errors -- we have instructions for onboarding journalists later, let's not duplicate them here.

## Checklist (Optional)


conorsch (approved):
Looks good, nice catch!

PR #275, opened 2021-11-16
Add link to preflight testing docs

conorsch:
## Status

Ready for review

## Description of Changes
As requested in [0], we now have detailed docs in the SD wiki for
developers, and there's an obviously appropriate place to link out from
the RM docs, so here we are.

[0] https://github.com/freedomofpress/securedrop/wiki/QA-Procedures#preflight-testing

## Testing
Render locally and make sure the link works as expeted. 

## Release 
Dev-only.

### DIFF
--- a/docs/development/release_management.rst
+++ b/docs/development/release_management.rst
@@ -286,9 +286,8 @@ Release Process
 #. A reviewer must verify the build logs, obtain and sign the generated ``Release``
    file, and append the detached signature to the PR. The PR should remain in
    draft mode. The packages on ``apt-qa.freedom.press`` are now installable.
-#. Coordinate with one or more team members to confirm a successful
-   clean install in production VMs using the packages on
-   ``apt-qa.freedom.press``.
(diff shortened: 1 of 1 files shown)


docs/development/release_management.rst
```diff
@@ -286,9 +286,9 @@ Release Process
 #. A reviewer must verify the build logs, obtain and sign the generated ``Release``
    file, and append the detached signature to the PR. The PR should remain in
    draft mode. The packages on ``apt-qa.freedom.press`` are now installable.
-#. Coordinate with one or more team members to confirm a successful
-   clean install in production VMs using the packages on
-   ``apt-qa.freedom.press``.
+#. Coordinate with one or more team members to `confirm a successful clean install in production VMs
+   <https://github.com/freedomofpress/securedrop/wiki/QA-Procedures#user-content-preflight-testing>`__
```
eloquence, line 290: Turns out this change to the anchor ID is necessary because GitHub only pretends that its anchors are anchors -- they're actually resolved via JavaScript, and the underlying anchor ID is hidden from view. (As a side effect, trying to visit an anchored GitHub URL without JavaScript means -- no working anchor. Nice one, GitHub!)

eloquence (approved):

PR #277, opened 2021-11-18
Fix inaccuracies in sample privacy policy

eloquence:
## Status
Ready for review 

## Description of Changes

1) The `last_updated` timestamp for sources is not removed or updated if a journalist deletes files or messages.

2) "Deleting" a reply as a source does not actually cause the file to be deleted. In fact, the Journalist Interface treats it as a  read receipt. (This is very counterintuitive behavior that we IMO should revise in future, but that's how it works right now.)

### DIFF
--- a/docs/deployment/sample_privacy_policy.rst
+++ b/docs/deployment/sample_privacy_policy.rst
@@ -27,13 +27,11 @@ Collection of Information From Sources
   message is automatically deleted.
 
 * Journalists decrypt and read each message offline. They are encouraged to
-  delete messages from the server on a regular basis. The date and time of any
-  message will be securely deleted from the server when the message is deleted.
+  delete messages from the server on a regular basis.
 
 * Please keep in mind that the actual messages you send and receive through
   SecureDrop may include personally identifying information. For this reason,
-  once you read a journalist’s message, we recommend you delete it. It will
-  then be securely deleted from the file system.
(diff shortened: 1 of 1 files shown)


docs/deployment/sample_privacy_policy.rst
```diff
@@ -27,13 +27,11 @@ Collection of Information From Sources
   message is automatically deleted.
 
 * Journalists decrypt and read each message offline. They are encouraged to
-  delete messages from the server on a regular basis. The date and time of any
-  message will be securely deleted from the server when the message is deleted.
+  delete messages from the server on a regular basis
```
zenmonkeykstop, line 30: nit: missing a period. 
optional change: it would be good to speak to the secure deletion part here. "Messages will be securely deleted from the server." or similar.

zenmonkeykstop (changes requested):
LGTM subject to de-nitting and possible optional change.

docs/deployment/sample_privacy_policy.rst
eloquence: Fixed. I've left the optional change for now to land this fix; might want to kick "secure delete" language around a bit more given different hardware configurations (SSD vs. non-SSD) but it's possible I'm overthinking it. :)

cfm (approved):
Approving with nit fixed, leaving the overthinking for another day.  :-)

PR #300, opened 2022-01-13
Bump copyright year

eloquence:
## Status

Ready for review. Do not merge - will be used to test docs deployment pipeline.


conorsch (approved):
Approving, but not merging, so we can time the deploy carefully.

PR #309, opened 2022-02-10
Make it clear that docs are now self-hosted

eloquence:
## Status

Ready for review

## Description of Changes

- We no longer host the docs via RTD as of earlier this year
- Fixed links pointing to RTD instance of support docs (let's separately investigate if we can redirect those readthedocs builds so folks don't accidentally land there)

## Checklist (Optional)


eloquence:
Thanks, both of those issues should now be addressed (I spun up a local build to test the updated config).

conorsch (approved):

PR #310, opened 2022-02-10
Use direct SSH for prod VMs

conorsch:
## Status

Ready for review 

## Description of Changes
In the upgrade testing docs, we were recommending ssh access on the VM
host, via molecule. That won't work, since prod VMs are fully firewalled
and drop all inbound. Instead, one should use the admin Tails VM, with
its custom SSH aliases configured by `./securedrop-admin tailsconfig`.

Encountered this during QA for 2.2.0. 

## Testing
Visual review is A-OK.

## Release
N/A, dev-only.

### DIFF
--- a/docs/development/upgrade_testing.rst
+++ b/docs/development/upgrade_testing.rst
@@ -58,7 +58,7 @@ To confirm:
 
 .. code:: sh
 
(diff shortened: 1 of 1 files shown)


eloquence (approved):
Makes sense, thanks!

PR #260, opened 2021-10-14
Add 2.1.0 docs; remove Focal/v3 docs; bump version

eloquence:
## Status

Ready for review

## Description of Changes

- Adds standard upgrade guide for 2.1.0. Simplifies language re: Tails updates given that this release attempts to correct the issue automatically. 
- Removes Focal migration guide and v3 guide. Folks who have not migrated at this point will likely need to start from scratch, but we still are retaining some v2 recovery docs in the backup/restore guide.
- Bumps version to 2.1.0; therefore in draft until day of release to avoid accidental merge.

## Checklist (Optional)

- [x] Doc linting (`make docs-lint`) passed locally
- [ ] Doc link linting (`make docs-linkcheck`) passed
- [x] You have previewed (`make docs`) docs at http://localhost:8000

PR #311, opened 2022-02-15
Add 2.1.0 to 2.2.0 upgrade guide

legoktm:
## Status

Ready for review

## Description of Changes

* Description: Add the 2.1.0 to 2.2.0 upgrade guide. Contents are mostly copied from the 2.0.2 to 2.1.0 version, with the Kernel part copied from the 2.0.1 to 2.0.2 version.

## Testing
* Read it :)

## Release 
* Should be coordinated with 2.2.0 release.

## Checklist (Optional)

  - it's hanging for me, I think it's on my end though

### DIFF
--- a/docs/upgrade/2.1.0_to_2.2.0.rst
+++ b/docs/upgrade/2.1.0_to_2.2.0.rst
@@ -1,12 +1,14 @@
-Upgrade from 2.0.2 to 2.1.0
+.. _latest_upgrade_guide:
+
+Upgrade from 2.1.0 to 2.2.0
 ===========================
 
-Updating Servers to SecureDrop 2.1.0
+Updating Servers to SecureDrop 2.2.0
 ------------------------------------
 Servers running Ubuntu 20.04 will be updated to the latest version of SecureDrop
 automatically within 24 hours of the release.
 
-Updating Workstations to SecureDrop 2.1.0
+Updating Workstations to SecureDrop 2.2.0
 -----------------------------------------
 
 .. note::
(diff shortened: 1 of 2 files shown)


legoktm:
I was reviewing https://github.com/freedomofpress/securedrop-docs/pull/260 and noticed that last time we updated the screenshots, apparently manually. Do we want to do that again for this release?

eloquence:
(It looks like `./update_version.sh` was broken in the RTD->self-hosting migration, I pushed a small commit to fix.)

eloquence:
Looks good to me! I'll convert this to draft for now to avoid accidental merges (even though it would only show up in `latest`, we want to preserve the 2.1.0 version in both for now to avoid users accidentally using the wrong version).

docs/upgrade/2.1.0_to_2.2.0.rst
```diff
@@ -0,0 +1,114 @@
...
+   upgrade prompt after connecting to the Internet, perform a
+   :ref:`manual update <Update Tails Manually>`. If this also fails, please
+   don't hesitate to contact us.
+
+Troubleshooting Kernel Issues
+-----------------------------
+SecureDrop 2.2.0 includes a kernel update on the *Application* and *Monitor
+Servers*, from version 5.4.136 to version 5.15.18. As with all kernel updates,
+we have extensively tested this update against
+:ref:`recommended hardware <Specific Hardware Recommendations>`.
```
eloquence, line 90: Assuming we move forward with the Mac Mini deprecation, we could insert another note to that effect here as well.

legoktm: Done.

eloquence (approved):

PR #312, opened 2022-02-15
Add deprecation notice for Mac Mini support

eloquence:
## Status

Ready for review

## Description of Changes
Assuming https://github.com/freedomofpress/securedrop/pull/6274 lands, we should be able to support the Mac Minis for one more round, but it's reasonable to expect a hardware upgrade at this point and to begin officially deprecating support. Leaving as draft for now, pending final confirmation that we can get them to work for 2.2.0.

## Checklist (Optional)

### DIFF
--- a/docs/hardware.rst
+++ b/docs/hardware.rst
@@ -515,11 +515,10 @@ instance we recommend using 7th- or 8th-generation NUCs instead.
 
 .. important::
 
-   SecureDrop 2.2.0 will be the last regular release of SecureDrop
-   with support for Mac Mini hardware, and this section of the documentation
-   will be removed from a future release. If you are still using Mac Mini servers,
-   you must move to a supported hardware platform as soon as possible to
-   continue using SecureDrop.
+   SecureDrop 2.2.x will be the last release series with support for Mac Mini
+   hardware, and this section of the documentation will be removed from a future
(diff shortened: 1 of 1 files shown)


docs/hardware.rst
```diff
@@ -513,17 +513,22 @@ instance we recommend using 7th- or 8th-generation NUCs instead.
 2014 Mac Minis
 ~~~~~~~~~~~~~~
 
+.. important::
+
+   SecureDrop 2.2.0 will be the last regular release of SecureDrop
```
legoktm, line 518: Question about versioning: Would we be dropping Mac Mini support in 2.2.1 then? Or would we do that in 2.3.0?

eloquence: I think we would default to attempting to continue support Mac Minis for any 2.2.x, while dropping official support for 2.3.0 (it may not be breaking just yet, but we would no longer test against that hardware). We could make that more explicit and say "The SecureDrop 2.2.x series" or similar.

legoktm: Your suggested wording sounds good to me!

legoktm (approved):
LGTM, not merging until the 2.2.0 release (I assume)

legoktm (approved):

PR #315, opened 2022-03-03
Update HTTPS certificate docs

conorsch:
## Status

Ready for review

## Description of Changes
Updates the docs on obtaining and/or renewing an HTTPS certificate via DigiCert. The process for requesting and renewing onion certs from DigiCert has changed substantially. These docs are based on feedback from SD admins, as well as from our own testing of the procedure via a recent request.

## Testing
Visual review is sufficient.

## Release 
No concerns.


conorsch:
Updating to ready-for-review. We're still waiting on a final cert from DigiCert based on a request we made a few weeks ago. I may follow up with additional clarifications once that process is complete, but for now, these clarifications are immediately useful to admins.

eloquence:
Thanks again for all your hard work on this, @conorsch! 

According to our most recent comms, DigiCert is now pointing folks to https://knowledge.digicert.com/generalinformation/onion-domains.html , which references its own `onionmaker` tool to generate the CSR.

I think we may want to confirm that this tool can be safely installed on the Admin Workstation, although it's possible that DigiCert will continue to accept CSRs generated by other means. Assuming we do want to test the DigiCert tool, I think this could potentially be picked up by another team member since we really just need to verify that it's safely installable on the AWS and can be used in a similar manner as the Harica tool.

docs/https_source_interface.rst
```diff
@@ -50,39 +50,60 @@ you will be required both to confirm your affiliation with the organization,
...
+Below are detailed steps for use on Tails:
 
 .. code:: sh
 
-    # From the Admin Workstation, SSH to the Application Server
-    $ ssh app
+    # On the Admin Workstation, generate the first CSR
+    $ mkdir ~/Persistent/sd-https-key-generation
+    $ cd ~/Persistent/sd-https-key-generation
+    $ openssl req -new -newkey rsa:4096 -nodes -keyout sd.key -out sd.csr
```
zenmonkeykstop, line 75: IMO it needs to be explicit that this is the key you'll be using with the final cert - anecdotally, because of the second CSR generated for domain validation purposes, this has been a point of confusion. So wording like "Save this key, as you will need it to use the certificate issued by Digicert" or something would be a good addition.

docs/https_source_interface.rst
```diff
@@ -50,39 +50,60 @@ you will be required both to confirm your affiliation with the organization,
...
+Interface, you will need to perform a signing operation leveraging the
+private key of the Onion service used on the Source Interface.
+DigiCert will provide you with some text and request that you use that text
+in a signing operation. At a high level, obtaining a certificate from DigiCert
+involves:
+
+1. Generating an HTTPS keypair and CSR via ``openssl``.
+2. Submitting the CSR to DigiCert.
+3. Scheduling a phone call and verifying your relationship to the organization.
+4. Generating another CSR, using a custom tool, leveraging the Onion service private key.
```
zenmonkeykstop, line 62: Docs should say what this CSR is for, vs the first one.

conorsch:
@zenmonkeykstop I've pushed a commit clarifying the need for two (2) CSRs. Please have another look. I'll switch over to looking at the `onionmaker` tool and follow up with changes in a separate PR.

conorsch:
> I'll switch over to looking at the onionmaker tool and follow up with changes in a separate PR.

Instead, I've pushed another commit to this PR. Ready for review.

zenmonkeykstop (approved):
Changes LGTM

PR #320, opened 2022-03-18
Update instructions on how we manage Tor packages

legoktm:
## Status

Ready for review

## Description of Changes

* Remove `fetch-tor-packages` from release management instructions,
  we are no longer tying it to releases, and the molecule playbook
  is obsolete and being removed.
* Add a new page that explains the CI process, what to test for new
  versions, and how to promote good releases.

Refs <https://github.com/freedomofpress/securedrop/issues/6233>.

## Testing

* Read instructions, verify they make sense

## Release 

* No special considerations

## Checklist (Optional)

### DIFF
--- a/docs/development/updating_tor.rst
+++ b/docs/development/updating_tor.rst
@@ -1,0 +1,36 @@
+Updating Tor
+============
+
+Given SecureDrop's significant reliance on Tor via Onion Services, we
+test new Tor versions to ensure they don't break SecureDrop before releasing
+them to users.
+
+Identifying new releases
+------------------------
+
+Announcements for new Tor releases are posted in the `Tor forum
+<https://forum.torproject.net/c/news/tor-release-announcement/28>`_.
+
+Our continuous integration automatically checks for new Tor packages every
+night and should commit them to the `securedrop-dev-packages-lfs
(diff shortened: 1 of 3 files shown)


docs/development/updating_tor.rst
```diff
@@ -0,0 +1,32 @@
...
+Our continuous integration automatically checks for new Tor packages every
+night and should commit them to the `securedrop-dev-packages-lfs
+<https://github.com/freedomofpress/securedrop-dev-packages-lfs>`_ repository.
+Within 15 minutes they should be available for download via
+``apt-test.freedom.press``.
+
+Testing
+-------
+
+Use a staging environment to verify that with the new Tor release, SecureDrop
```
zenmonkeykstop, line 23: Ideally, a prod environment should be used and left to run through the unattended-upgrade cycle, to confirm that Tor will still be on the correct version and running as expected after a reboot.

zenmonkeykstop (changes requested):
Docs look good, but would prefer that the Tor debs be checked on a prod setup overnight...

docs/development/updating_tor.rst
legoktm: Added, mostly using your wording here :)

PR #345, opened 2022-04-07
Use graphical Disks app to create Ubuntu USB instead of `dd`

### LINKED ISSUE #338: Tell people to use GNOME Disks to create Ubuntu installer USB
## Describe the change
https://docs.securedrop.org/en/stable/servers.html#create-the-ubuntu-installation-media

Using `dd` is error-prone if you grab the wrong `/dev/sdX` ID and requires usage of the CLI. GNOME Disks makes it like 3 clicks to restore the ISO onto a USB, and is also what Tails recommends: https://tails.boum.org/install/linux/index.en.html

* Guide admins into using a GUI program that is less error-prone than running a CLI command

legoktm:
## Status

Ready for review

## Description of Changes

`dd` requires more command-line familiarity and is way too easy
to get wrong if you accidentally pick the wrong USB device letter.

We already use Disks earlier in the process to set up the Transfer
Device, plus it's what Tails recommends, so let's use it here too.

Includes guided screenshots to make the process as straightforward
as possible.

Fixes #338.

## Testing

* Follow the new instructions

## Release 

* n/a

## Checklist (Optional)


docs/servers.rst
```diff
@@ -136,28 +136,25 @@ Create the Ubuntu Installation Media
...
-easy to identify which USB has which sdX identifier. If you are unsure, try
-running ``lsblk`` before and after plugging in the USB you are using for the
-Ubuntu installer. Note that you should use the main block device (e.g. ``/dev/sdb``)
-rather than any listed partitions (e.g. ``/dev/sdb2``).
-
-If your USB is mapped to /dev/sdX and you are currently in the directory that
-contains the Ubuntu ISO, you would use dd like so: ::
-
-   sudo dd conv=fdatasync if=ubuntu-20.04.4-live-server-amd64.iso of=/dev/sdX
+CD-R or create a bootable USB stick. Open the **Disks** program
```
eloquence, line 139: Maybe omit CD-R entirely or say "To create a bootable USB stick, open .."

docs/servers.rst
```diff
@@ -136,28 +136,25 @@ Create the Ubuntu Installation Media
...
+(**Applications ▸ Utilities ▸ Disks**) used earlier.
+
+Select the USB device you'd like to use and then open the menu in the top right,
```
eloquence, line 142: Soft stylistic preference for "you would" over "you'd" for documentation.

docs/servers.rst
```diff
@@ -136,28 +136,25 @@ Create the Ubuntu Installation Media
...
+selecting **Restore Disk Image**.
+
+|screenshot of restore disk image|
+
+Select the Ubuntu ``.iso`` file downloaded earlier and click **Start Restoring**.
+
+|screenshot of start restoring|
+
+Once it finishes, you can remove USB stick as it's ready to use.
```
eloquence, line 151: As above, soft preference for "it is" over "it's".

eloquence (commented):
Nice change, a couple of small nits. Leaving open a few days in case anyone sees a reason to still mention `dd` or prefer it.

docs/servers.rst
legoktm: Did both.

docs/servers.rst
legoktm: Done.

docs/servers.rst
legoktm: Done.

docs/servers.rst
```diff
@@ -135,29 +135,25 @@ following output in your terminal. ::
...
-
-If your USB is mapped to /dev/sdX and you are currently in the directory that
-contains the Ubuntu ISO, you would use dd like so: ::
-
-   sudo dd conv=fdatasync if=ubuntu-20.04.4-live-server-amd64.iso of=/dev/sdX
+To create the Ubuntu installation media on a bootable USB stick, open the
+**Disks** program (**Applications ▸ Utilities ▸ Disks**) used earlier.
+
+Select the USB device you would like to use and then open the menu in the top
+right, selecting **Restore Disk Image**.
```
eloquence, line 142: As discussed in sprint planning yesterday, it may be appropriate to add a brief reminder here that this list includes all disks, so it's important to ensure that the selection is correct to avoid accidentally overwriting data.

legoktm: Done.

eloquence (approved):

Issue #332, opened 2022-04-06
Explain how to generate a diceware passphrase


legoktm:
## Describe the change
AFAICT there's no place where we actually describe how to use KeePassXC to generate a diceware passphrase. https://docs.securedrop.org/en/stable/set_up_tails.html#enable-persistent-storage (the first time you need to generate one), links to a 2015 article from The Intercept, which explains one way to do it, but using KeePassXC's built-in passphrase generator is easier, especially we later on use KeePassXC to store other passphrases.

This explanation should include how to generate the passphrase, and what parameters to use (e.g. is the default of 7 words ideal?)

## How will this impact users?
* Ensure that admins can quickly generate strong passphrases

Issue #335, opened 2022-04-06
Provide guidance on how strong the Tails admin password should be


legoktm:
## Describe the change
When using the admin workstation, we enable a root Administrative password in Tails. This password is only used for that Tails session (c.f. https://docs.securedrop.org/en/stable/set_up_admin_tails.html#start-tails-with-persistence-enabled "The Tails administration password is a one-time password. It will reset every time you shut down Tails.").

How strong do I need to make this password? Does it need to be a full diceware passphrase or can it be as simple as "password"? I don't fully understand the threat model of this password.

PR #351, opened 2022-05-05
Permits option to create a diceware passphrase using KeePassXC

sts10:
## Status

Ready for review 

## Description of Changes

* Adds more specific advice on how to generate Diceware passphrases to the "Enable Persistent Storage" section of the "Create Tails USBs" page. As noted in #332, the current version of this section rather opaquely advises "We recommend using Diceware passphrases", with a link to a (7-year-old) Intercept article that primarily instructs readers to use physical dice to generate passphrases. If we believe that is the level of security needed in this case, than we should be clear that users should use the physical-dice method. But assuming that we trust KeePassXC, I took the liberty of adding KeePassXC's passphrase generator as an alternate option. I included a screenshot from KeePassXC v 2.6.6. 
* There's also a question of how many words the passphrase should be. I assumed 7 (~90 bits) is fine, but I defer to sharper minds who better understand the threat model here. But I do think we should give a number. I see #335 asking this question.

* Fixes Issue #332

## Testing

* I set up the Sphinx environment and ran `make docs`. Looks good.

## Checklist (Optional)

- [X] Doc linting (`make docs-lint`) passed locally
- [X] Doc link linting (`make docs-linkcheck`) passed
- [X] You have previewed (`make docs`) docs at http://localhost:8000


sts10:
We might also want to switch the external link from that Intercept article to [this EFF page](https://www.eff.org/dice), which gets to the actual steps of using physical dice to generate passphrases a bit quicker.

sts10:
Ah, now realizing we probably should give passphrase-generating instructions all in one place, likely on the "Passphrase Best Practices" page, and then link to that from sections like "Enable Persistent Storage". I'll close this PR now and work on that. I welcome comments though!

eloquence:
@sts10 Thanks so much for working on this! Looks great to me in principle, have also flagged to our DigiSec team in case they want to weigh in on DiceWare best practices and KeePassXC's implementation in particular. I personally do use the KeePassXC diceware functionality these days, but I've not looked closely at its underlying dictionary implementation.

huertanix:
I haven't yet seen solid evidence showing that 7 words is necessary for Tails persistent storage passphrases. Most password cracking models that I've explored don't seem to go into detail on whether or not they take things like the speed of a hashing function into account. If LUKS's encryption scheme would have any mitigations that would slow password cracking attempts, as algorithm implementations in some password managers do, for example (https://security.stackexchange.com/a/137307, "Bruteforcing the master password"), then it may be that a shorter passphrase would be fine, and I think it's useful to explore that possibility.

Based only on my anecdotal evidence (but in this case anecdotes from newsrooms) the longer a passphrase is, the more difficult it will be to memorize. Without going into detail I've seen journalists use other means to store their long passphrases. Easier-to-remember passphrases may make those other means less necessary.

sts10:
> If LUKS's encryption scheme would have any mitigations that would slow password cracking attempts, as algorithm implementations in some password managers do, for example (https://security.stackexchange.com/a/137307, "Bruteforcing the master password"), then it may be that a shorter passphrase would be fine, and I think it's useful to explore that possibility.

I took this as a challenge to do a little investigating. I installed and started up Tails (first time!) following [these instructions](https://tails.boum.org/install/linux/index.en.html). I then created a persistent storage partition (using a trivial passphrase) following [these instructions](https://tails.boum.org/doc/first_steps/persistence/index.en.html). In both cases, I was attempting to do what I'd assume a regular SecureDrop user would do, following current documentation. 

Then I went off-script, and I set an admin password for Tails and installed cryptsetup from [this tar file](https://gitlab.com/cryptsetup/cryptsetup) (installing via `sudo apt install cryptsetup` was giving me issues, I assume because its Tails). I then ran `sudo ./cryptsetup luksDump /dev/sdb2` to learn more about our new encrypted partition. Here's what it output:

IMG_3621

I _think_ the key lines for our question are "Hash spec" (sha256) and "MK iterations" (113580). I believe the former means that the partition uses PBKDF2-sha256 to hash the passphrase, using ~113k iterations.

I'm guessing the next logical step might be to run [hashcat](https://hashcat.net/hashcat/) against these settings and see how long it takes to break 7776^4, 7776^5, 7776^6, etc. passphrases, given some sane hardware assumptions? Hope this moves the discussion forward!

sts10:
Some back-of-the-envelope math, just for fun:

Running the wonderful `cryptsetup benchmark` command on my Oryx Pro informs me that, against PBKDF2-sha256, it gets `1736052 iterations per second for 256-bit key`.

Given that my test key requires 1,820,444 iterations, that's about 1.05 second per attempt (very curious if Tails somehow used my laptop to benchmark that 1.8M number?). `1.0486 * 7776**3 /60 / 60 / 24 / 365 / 100` gives us about 156 centuries to get through the entire space for a 3-word passphrase. 4 words gets us to 121,500 millennia.

eloquence:
@sts10 Thanks for your patience, and thanks again for this PR. Upon consideration, I agree with your earlier point of consolidating this passphrase generation information in the passphrases section.

https://docs.securedrop.org/en/stable/passphrases.html

I also agree that we should link to the EFF guide instead of The Intercept article. The Intercept has started adding a login wall to articles, which also means that we can no longer consider their content to be easily accessible. 

Do you have time to iterate on this PR? If not, I'd be happy to do so!

sts10:
Here's an attempt at said consolidation. Hope I got the Sphinx linking correct... I've changed distros since I wrote this PR and `sphinx-autobuild` isn't cooperating this go-around.

> I haven't yet seen solid evidence showing that 7 words is necessary for Tails persistent storage passphrases. 

I kept my initial, conservative recommendation of 7 words in this PR edit, though it sounds like we can knock that down to 5 or even 4? 

Also, I see now that I got an extra commit (c1492da) in there when I clicked GitHub's relatively new "Fetch Upstream" button... my git powers aren't fantastic. Let me know if/how I can clean that up or if it's best for me to just start a fresh PR.

eloquence:
Thanks much, should be able to take another spin tomorrow! Let's indeed stick to the more conservative rec for now to give more folks a chance to weigh in on that independently, but that needn't hold up the recommendation to use a convenient tool (which uses the EFF wordlist under the hood). :)

eloquence (approved):

eloquence:
Thanks @sts10! I've taken the liberty to squash the commits. I generally recommend practicing interactive `git rebase` commands -- it's a bit tricky to master, but very handy once you get the hang of it. You can use it both to keep a repository in sync with upstream (see, e.g., https://ljvmiranda921.github.io/notebook/2021/11/12/sync-your-fork/ for a simple tutorial) and to eliminate commits you don't need to keep in the history. That said, the merge commits GitHub's web-based workflow generates are not hugely troublesome either way.

Thanks again for working on this!

PR #360, opened 2022-07-12
Revert "Add instructions for manually downloading Tor signing key"

eloquence:
Reverts freedomofpress/securedrop-docs#359

The workaround is not required with recent Tails versions.

Since this is documentation for a fresh install, we can reasonably assume that users will be using the most recent release of Tails.  In fact, this was included in Tails 5.1.1, released on 2022-06-23 (although I don't see it mentioned in the changelog, I've verified via `git tag --contains` that the relevant commit https://gitlab.tails.boum.org/tails/tails/-/commit/f313d415e1e49f79b26ac70cd55da94808766a40 is included in both 5.2 and 5.1.1).

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -5,26 +5,9 @@ Install Prerequisites
 ---------------------
 
 SecureDrop has dependencies that need to be loaded onto the *Admin Workstation*
-before installing the servers.
-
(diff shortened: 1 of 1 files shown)


cfm (approved):
I can confirm from testing SecureDrop v2.4.1 from Tails 5.2 that this workaround is no longer required.

PR #362, opened 2022-07-20
Fix Qubes RPC policies to match R4.1 requirements

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

Updates the Qubes RPC policies to match the Qubes OS R4.1 requirements.

## Testing

- [ ] Confirm that `make staging` fails before writing the RPC policies.
- [ ] Confirm that `make staging` succeeds after writing the RPC policies.

## Release 

* No special considerations for release.

## Checklist (Optional)

### DIFF
--- a/docs/development/qubes_staging.rst
+++ b/docs/development/qubes_staging.rst
@@ -5,7 +5,7 @@ SecureDrop currently uses Ubuntu Focal as its server OS.
 The instructions below cover setting up a SecureDrop staging environment
 using Focal under Qubes.
 
-It is assumed that you have an up-to-date Qubes installation on a compatible
+It is assumed that you have an up-to-date Qubes R4.1 installation on a compatible
 laptop, with at least 16GB RAM and 60GB free disk space. The SecureDrop server VMs
 run Tor locally instead of using ``sys-whonix``, so the system clock must be set
 accurately for Tor to start and hidden services to be available.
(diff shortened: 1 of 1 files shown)


zenmonkeykstop (changes requested):
Just worked through setting up a 4.1 staging env with this, looks good! (There were a few RPC denies that did not affect the provisioning process.)

One other point that might be adding (tho it's not a policies issue, just a 4.1 one) would be that if you accept the default dispvm setup for `sys-firewall`, you have to add the iptables script in its templateVM (`fedora-*-dvm` by default) rather than in `sys-firewall` itself.

gonzalo-bulnes:
Thanks for the review @zenmonkeykstop ! I'll ask you about that as I haven't encountered the scenario. It seems to make sense to add that here indeed.

gonzalo-bulnes:
@zenmonkeykstop 

1. Should we include policies to avoid noisy access denials even if they don't prevent the provisioning to succeed? (I'm inclined to think that we should, because I feel a little bit sad every time I follow logs that contain avoidable errors, warnings especially if they're unrelated to what I'm troubleshooting.)
2. would you mind tracking the iptables setup changes in a separate issue? (For the sake of keeping this one scoped.)

rocodes:
As for @zenmonkeykstop's comment: indeed the firewall rules have to be added to (disp) sys-firewall's template, otherwise the rule will disappear whenever the user reboots sys-firewall, and stop them from sshing into the staging servers each time until they re-add the rule.

Our Salt states currently enforce `fedora-36-dvm` as the template (disp) for sys-firewall when using the workstation, so the path of least resistance for developers is to ask them to modify that template to include the staging firewall rules.

This family of change is a _small_ potential gotcha with regards to debugging (in that a system that was staging and is flipped to prod will not be identical to a prod system, because the contents of the VMs aren't checked), but we kind of know that already.

zenmonkeykstop (approved):

PR #364, opened 2022-07-20
document workaround for DHCP collision in `bento/ubuntu-20.04` v202206.03.0

cfm:
## Status

Ready for review

## Description of Changes

Bringing up the `app` and `prod` VMs will fail using `bento/ubuntu-20.04`
v202206.03.0 (freedomofpress/securedrop#6497), so here we pin to the
last-working v202111.19.0.

This can be reverted once chef/bento#1421 is resolved.

## Testing

* [ ] Changes look good on visual review.
* [x] *Optional:* Revised `vagrant box add` command correctly pulls the intended version.

## Release 

No special considerations; development-only.

## Checklist (Optional)

### DIFF
--- a/docs/development/virtual_environments.rst
+++ b/docs/development/virtual_environments.rst
@@ -214,11 +214,16 @@ Set the default Vagrant provider to ``libvirt``:
 
 Convert Vagrant boxes to libvirt
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-Convert the VirtualBox images for Focal from ``virtualbox`` to ``libvirt`` format:
+..
+   See also: securedrop#6497
+
+Convert the VirtualBox image for Focal from ``virtualbox`` to ``libvirt`` format.
+Pending resolution of an `upstream bug <https://github.com/chef/bento/issues/1421>`_,
+we must use an older version of this image.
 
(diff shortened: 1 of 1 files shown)


nathandyer (changes requested):
Everything looks correct, but with the split of the developer docs out into its own repo, this is no longer where this change should live.

I know it's possible to transfer issues to a new repo, but I don't know of a way to transfer a branch or a PR in a way that doesn't break everything.

@cfm would you mind to open a PR in the `securedrop-dev-docs` repo with these changes? I can also copy this over and create a new one myself if you prefer, but I thought it would be best to preserve original authorship.

cfm:
Thanks for reviewing, @nathandyer.  Let me check today (alongside QA) whether freedomofpress/securedrop#6497 and the underlying chef/bento#1421 are still true after chef/bento#1429, and then I'll either re-open this pull request in `securedrop-dev-docs` or just close it out entirely.

cfm:
Testing against a new Ubuntu 20.04.5 image appears to be blocked on <https://github.com/chef/bento/issues/1405#issuecomment-1125876976>, so I've gone ahead and refiled this workaround in freedomofpress/securedrop-dev-docs#19.

Issue #347, opened 2022-04-19
Update documentation to reflect changes in Tails 5.0, when available


zenmonkeykstop:
## Describe the change

Tails 5.0 includes some UI changes as a result of the move to Debian 11. Docs and screenshots should be updated to match once Tails 5.0 is released (provisionally scheduled for mid-late May).

DrWhax:
FWIW, i'm going through one right now:

- There's no settings button anymore under the top-right menu.
- Kleopatra is the new application to export PGP keys from.

eloquence:
(Not fully resolved yet, some Journalist Guide updates pending.)

PR #371, opened 2022-08-10
Added new screenshots and documentation for Submission Key generation in Tails 5.0

nathandyer:
## Status

Ready for review

## Description of Changes

* Description:
    * Changed the documentation to reflect the new workflow in Tails 5.0 related to Kleopatra for exporting the Submission Key and fingerprint.
    * Removed one [now] unnecessary screenshot
    * Updated the other three screenshots for the new workflow in Kleopatra
    * Added some clarifying details about the fingerprint

* Partially Fixes Issue #347 (Not Fully Resolved)

## Testing
* Changes should be visible as soon as the docs are built using the standard commands. You can check the changes using this address once the local build is running: http://127.0.0.1:8000/generate_submission_key.html 

## Checklist (Optional)

- [ x ] You have previewed (`make docs`) docs at http://localhost:8000


docs/generate_submission_key.rst
```diff
@@ -56,36 +56,36 @@ Create the Key
 Export the *Submission Public Key*
 ----------------------------------
 
-Navigate to **Applications ▸ Utilities ▸ Passwords and Keys** to open a
-graphical interface to manage GPG keys. Click the clipboard icon |gpgApplet|
-in the top right corner and select "Manage Keys". Click "GnuPG keys" and you
-should see the key that you just generated.
+Navigate to **Applications ▸ Utilities ▸ Kleopatra** to open a
```
eloquence, line 59: This is under "Accessories" for me at least on Tails 5.2.

docs/generate_submission_key.rst
```diff
@@ -56,36 +56,36 @@ Create the Key
...
+graphical interface to manage GPG keys. At the bottom of the list
+of keys, you will find the SecureDrop Submission Key you just created.
```
eloquence, line 61: It showed up under the topmost item in the list for me, but that may be because multiple keys were already present in the keyring. To avoid any situation where this is inaccurate, I would just say "In the list of keys .." (or instruct them to sort the list of keys first).

eloquence (changes requested):
Two nits but otherwise LGTM!

docs/generate_submission_key.rst
nathandyer: I double-checked, and you are correct about it not being in Utilities any longer. Apparently it can be found in both "Accessories" as well as "Favorites," which is shown by default when you open the Applications menu. Since the Favorites list can change, I'll specify "Accessories" here since that is consistently where it can be found.

docs/generate_submission_key.rst
nathandyer: Thank you, I (incorrectly) assumed that new keys would be added to the bottom since that's what I've seen each time I've followed these steps. Just to avoid any discrepancies based on how users sort their list of keys, I'll change this to just say that the SD Submission Key can be found in the list.

eloquence (approved):
LGTM, thank you!

Issue #291, opened 2021-12-16
Update Tails screenshots featuring the Applications > System Tools menu


zenmonkeykstop:
## Describe the change

The addition of the backup utility in 4.25 has changed the System Tools menu: it includes said utility now, and (annoyingly) it now has a scrollbar, with the user needing to scroll to bring the Terminal item into full view.

Since we refer to said item, and we wanna make sure users don't choose the Root Terminal item instead, it would be good to update screenshots illustrating it.

## How will this impact users?

Reduce opportunity for user error thru executing stuff as root

## User Research Evidence

nada

## Additional context

eloquence:
Resolved via #372

PR #372, opened 2022-08-12
Update remaining documentation with Tails 5 screenshots and changes

### LINKED ISSUE #347: Update documentation to reflect changes in Tails 5.0, when available

nathandyer:
## Status

Ready for review

## Description of Changes

* Updated the documentation for the following pages to bring the documentation up-to-date with changes to Tails 5:
    * Journalist Guide
    * Set up Transfer and Export Device
    * Backup Workstation
    * Offboarding
    * Printer Setup in Tails

* Resolves #347 

## Testing

* Build the docs repo using the steps in the README, then review the pages mentioned above.
* **Special note**: most of the cleanup was just adding in new screenshots, but I made significant changes to the Offboarding and Printer Setup documentation. Please review the textual changes there, and make sure I have not overlooked anything or eliminated anything that is still applicable.

## Checklist (Optional)


eloquence:
(Pushed a small commit to remove the "in Tails 4.20" qualifier for the Tor Connection Assistant, since Tails 4.20 is more than a year old and we can IMO reasonably assume that journalist users will encounter the TCA.)

docs/offboarding.rst
```diff
@@ -166,42 +166,39 @@ was in effect.
...
-#. Double-click the key, and in the **Names and Signature** tab, add a name
-   such as "OLD <Your Organization> SecureDrop Submission Key - Do Not Delete
-   - Retired <Date>". (This is a local-only change to stop you from mixing up
-   the old and new keys).
-
+   |add email|
 
+#. Change the name of the key to "OLD <Your Organization> SecureDrop Submission
+   Key - Do Not Delete", and set the comment to "Retired <Date of Retirement>".
+
```
eloquence, line 186: From my testing, this adds a UID rather than changing it (though it pre-populates the fields with information from the existing UID). I also don't see a better option in Kleopatra, and it seems to do the job in setting the key apart. 

I'd suggest a slight reword for clarity:

> Set the name field to "OLD <Your Organization> SecureDrop Submission  Key - Do Not Delete", and set the comment field to "Retired `<Date of Retirement>`". Click **OK** to add this information to the key.

docs/offboarding.rst
```diff
@@ -250,16 +247,7 @@ Return to the Secure Viewing Station
 
 #. **Do not delete your old submission key!** You'll want to maintain it on
    the *SVS* so that you can still decrypt old submissions that were made
-   before you changed keys. If you like, you can revoke the key by
-   selecting the key in the **Passwords and Keys** application, opening the
-   **Details** tab, highlighting the first key entry and clicking
-   **Revoke**. This also makes local-only changes and does not stop you or
-   anyone else from using the key, but it is a reminder that your key has
-   changed.
-
```
eloquence, line 259: The way to do this in Kleopatra appears to be to generate a revocation certificate, edit it with a text editor to remove the colon inserted for security reasons, and then import it into Kleopatra.

eloquence (changes requested):
Looks great, a couple of changes suggested in the offboarding guide. I've also pushed two minor commits - please take a look at those and give me a :+1: to indicate if you're cool with those tweaks. :)

docs/offboarding.rst
nathandyer: @eloquence Thanks! I'm trying to document this process, but I can't seem to get Kleopatra to actually save the revocation certificate. Have tried saving in various places in the filesystem, as well as the Persistent storage, on multiple Tails USBs. Just to note here, the version of Kleopatra on the latest Tails is 3.1.11.

After digging into this a bit further, I think I'm running into this bug: https://bugs.kde.org/show_bug.cgi?id=384013

When you tested this, was it using a local copy of Kleopatra, or was it booted to a Tails disk? Do you happen to know the version? 

In the meantime, I'm going to try and come up with instructions using the CLI to avoid these bugs.

nathandyer: I just pushed documentation for a process that uses the CLI to create the revocation certificate to work around the bug mentioned above. Please review and let me know if you think this is a good method to avoid the issues I ran into earlier. Thanks!

nathandyer:
I believe this also satisfies issue #291

docs/offboarding.rst
eloquence: I tested it in Tails, and I think I know why we saw different behavior. I started `kleopatra` from the terminal first, where I see output related to the revocation, and it's saved correctly. If you start the application without a TTY attached to it, I am also unable to save it (it may never successfully shell out to `gpg`). I think CLI instructions are indeed the way to go, will review.

eloquence (approved):
Looks great and new revocation docs worked as expected! :shipit:

PR #384, opened 2022-09-15
Removed references to XML rulesets

### LINKED ISSUE #283: Remove all references to firewall XML rulesets
Currently, the Unsafe Browser that ships with Tails 4.24 intentionally blocks the ability to upload files from it (see: https://gitlab.tails.boum.org/tails/tails/-/issues/17284#note_1120). Because of this, there is currently no practical means of actually using said XML files, and their mention should be removed to prevent admins from having to discover this the hard way (yes, there's a warning that it's for reference only and yes, many admins will ignore that and try anyway). This will also help trim the docs a bit, which may make them slightly less overwhelming.

nathandyer:
## Status

Ready for review

## Description of Changes

* Description: Removes the references to XML rulesets for both firewall configuration guides. They were (mostly) already removed from the opnsense guide, and now the pfsense guide matches. This change was suggested in issue #283, because the Unsafe Browser does not allow uploads. 

* Resolves #283

## Testing

Build docs using steps in the README

## Checklist (Optional)

### DIFF
--- a/docs/firewall_opnsense.rst
+++ b/docs/firewall_opnsense.rst
@@ -385,7 +385,7 @@ Established", and the Tor Connection assistant should show the message "Connecte
 to Tor successfully".
 
 For the next step, SecureDrop Configuration, you will manually configure the
-firewall for SecureDrop, using screenshots or XML templates as a reference.
+firewall for SecureDrop, using screenshots as a reference.
 
 SecureDrop Configuration
 ------------------------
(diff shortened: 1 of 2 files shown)


eloquence (commented):
LGTM. @zenmonkeykstop Any thoughts from your end prior to merging this? I would suggest we also remove the XML files from the `securedrop` codebase if we have no intention of referencing or recommending them.

eloquence:
Briefly discussed with Kev, good to go :)

eloquence (approved):

PR #385, opened 2022-09-16
Add a labeled photo showing a full SD setup

### LINKED ISSUE #58: Include a picture of one full SD setup
## Description

Reporting this here while documenting TODOs (#724).

From `docs/overview.rst`

```rst
.. todo:: A picture of an actual physical setup (e.g. the office
          setup) with the components labeled would also be good here.
```

nathandyer:
## Status

Ready for review

## Description of Changes

* Description: Adds a photo, complete with labels, showing a full SecureDrop setup

* Resolves #58

## Testing

Make docs using steps in Readme, then review the 'Overview' page

## Checklist (Optional)


zenmonkeykstop (commented):
This has the potential to be misleading - best practice is to host the server components in an area to which only the admins have access, to store the SVS securely such that only journalists have access, and not to have networked devices in close proximity to the SVS (for airgap-jumping exploit reasons, though the odds of said exploits working in the wild are remote). I can see an argument for, for example, a top-down view of the servers and router all wired up, but I don't know that this photo really helps with anything.

nathandyer:
That's a very good point, and I agree. I think the photo helps to form a mental model if you're coming to this for the first time, but otherwise I don't think it's particularly helpful, and I'm not sure of how you could realistically pack everything into a photo or two that also is in line with our best practice recommendations.

I'm also not sure that a top-down view showing everything connected would be helpful, as it would be hard to tell what is connected to what, and would really just be a few boxes and a jumble of wires.

I think the thing to do might be to close both the PR and original issue, as there isn't much gained here. @zenmonkeykstop does that sound reasonable?

zenmonkeykstop:
Works for me.  Closing.

nathandyer:
Thanks for that perspective, @legoktm! If this is something that adds value, I can take a new photo only showing the app/mon servers, and the firewall. @zenmonkeykstop What are your thoughts?

Issue #210, opened 2021-04-30
Give example of onion name placement in landing page content suggestions


eloquence:
Now that more and more news orgs use an onion name for their SecureDrop, we should develop consistent recommended language that folks can show on their landing page to indicate both the onion name and the full-length v3 address.

nathandyer:
Resolved by https://github.com/freedomofpress/securedrop-docs/issues/184

Issue #37, opened 2020-10-20
Consider recommending HTTP/2 in landing page guidelines


eloquence:
The HTTP/2 spec [recommends](https://tools.ietf.org/html/rfc7540#section-9.2.2) the adoption of a cipher suite blacklist by implementers, clarifies security requirements for TLS connections, and includes other protocol-level security improvements. For security reasons, we should consider recommending it as part of our landing page guidelines.

zenmonkeykstop:
This still makes sense - given that landing pages may be delivered via a variety of means, specific instructions for Apache/Nginx/whatever would not be that helpful. Just calling it out as a recommendation should be enough.

PR #388, opened 2022-09-21
Landing Page Overhaul

### LINKED ISSUE #184: Improve Landing Page docs

nathandyer:
## Status

Ready for review

## Description of Changes

* Description: This PR makes significant changes to the landing page documentation by:
  * Explaining about the SecureDrop directory
  * Describing short onion addresses, and how to display them on a landing page
  * Explaining the submission process (instance verification request)
  * Marking the relevant, strictly enforced requirements for landing pages to be included in our directory
  * Adding `Permissions-Policy` headers to list of suggestions

* Resolves #184 , #210, and  #46 

* Related Issues
  * #37 (I couldn't find an elegant way to explain how to enable this in apache/nginx, but I mention it here as this would be the appropriate place for those changes as well)

## Testing

* These changes can be tested by building the branch according to the readme

## Checklist (Optional)

### DIFF
--- a/docs/deployment/landing_page.rst
+++ b/docs/deployment/landing_page.rst
@@ -24,17 +24,19 @@ the requirements below, then `send us a request using this form.
 There are several benefits to being included in the SecureDrop directory. The
 most significant benefit is that it will be easier for potential sources to
 find your SecureDrop instance. Additionally, being included in the directory
-makes you eligible for a short onion address. This improves the experience by
-turning a lengthy, non-descriptive address into one that is short and 
-memorable. For example, a long .onion address might look like: ::
+makes you eligible for `an onion 
+name. <https://securedrop.org/faq/getting-onion-name-your-securedrop/>`__
+This improves the experience by turning a lengthy, non-descriptive address
+into one that is short and memorable. For example, a long .onion address 
+might look like: ::
 
-    xp44cagis447k3lpb4wwhcqukix6cgqokbuys24vmxmbzmaq2gjvc2yd.onion
+    sdolvtfhatvsysc6l34d65ymdwxcujausv7k5jk4cy5ttzhjoi6fzvyd.onion
 
-whereas the short onion address might look like: ::
+whereas the shorter onion name might look like: ::
 
-     theguardian.securedrop.tor.onion
+    nyworld.securedrop.tor.onion
 
-If you wish to receive a short onion address, one can be provided during the
+If you wish to receive an onion name, one can be provided during the
 instance verification process. The format for short onion addresses is: ::
 
     organization.securedrop.tor.onion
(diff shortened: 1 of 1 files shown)


docs/deployment/landing_page.rst
```diff
@@ -15,6 +15,38 @@ your organization.
...
+find your SecureDrop instance. Additionally, being included in the directory
+makes you eligible for a short onion address. This improves the experience by
+turning a lengthy, non-descriptive address into one that is short and 
+memorable. For example, a long .onion address might look like: ::
+
+    xp44cagis447k3lpb4wwhcqukix6cgqokbuys24vmxmbzmaq2gjvc2yd.onion
+
+whereas the short onion address might look like: ::
+
+     theguardian.securedrop.tor.onion
```
eloquence, line 35: Let's pick a fictional example here (both to avoid accidentally having an old SD onion address in the docs at some future point, and as a rule of thumb to only reference external organizations with their explicit knowledge). We can use the same example as we did here: https://securedrop.org/faq/getting-onion-name-your-securedrop/#what-are-onion-names

docs/deployment/landing_page.rst
```diff
@@ -129,6 +171,19 @@ SecureDrop-related behavior.
 Instead, we recommend including .onion addresses in plain text, without a
 hyperlink.
 
+If you have been provided a short onion name for your instance, this address
+will also need to be plain text, without a hyperlink. We recommend using the
+text below to provide maximum clarity: ::
+
+    The SecureDrop instance can be found by entering the following address
+     in Tor browser: <short onion address>
```
eloquence, line 179: We generally recommend being explicit about being desktop vs. mobile (onion names won't work in the mobile app, and there are other reasons not to access SecureDrop from a mobile device), and we've standardized on the title case version of the application name. So I would say here: "the desktop version of Tor Browser".

docs/deployment/landing_page.rst
```diff
@@ -15,6 +15,38 @@ your organization.
...
+SecureDrop `maintains a directory of instances that meet our strict guidelines.
+<https://securedrop.org/directory/>`__ If you would like to be considered for
+inclusion in this directory, make sure your landing page is in compliance with
+the requirements below, then `send us a request using this form.
+<https://securedrop.org/directory/submit/>`__
+
+There are several benefits to being included in the SecureDrop directory. The
+most significant benefit is that it will be easier for potential sources to
+find your SecureDrop instance. Additionally, being included in the directory
+makes you eligible for a short onion address. This improves the experience by
```
eloquence, line 27: Tor has adopted the name "onion names" for this functionality, so I would suggest we standardize on that here as well. In case you have not seen it, we also have an [FAQ for onion names here](https://securedrop.org/faq/getting-onion-name-your-securedrop/), which we can cross-link here.

eloquence (changes requested):
Great work, thank you Nathan! A few small requested changes inline.

nathandyer:
Thanks for the review @eloquence! Making those changes led to a few other tweaks for the sake of consistency, so please let me know if I need to make any additional adjustments.

eloquence (approved):
LGTM!

PR #398, opened 2022-10-28
Flesh out QR code warning; add note about use of electronic devices in viewing area

nathandyer:
## Status

Ready for review 

## Description of Changes

* Description: This change adds a warning to the Journalist Guide, in the SVS section, to warn against scanning QR codes as described in [an earlier blog post](https://securedrop.org/news/security-advisory-do-not-scan-qr-codes-submitted-through-securedrop-connected-devices/). It also advises leaving any personal electronic devices outside of the secure area where the SVS is used.

## Testing
* Build according to the steps in the README and view the changes in the browser

## Checklist (Optional)

### DIFF
--- a/docs/journalist.rst
+++ b/docs/journalist.rst
@@ -569,9 +569,13 @@ are important steps you can take to protect yourself:
 
 .. warning::
 
-   For maximum safety, it is recommended that cell phones and personal 
-   electronic devices be left outside of the secure environment where the
-   *Secure Viewing Station* is used.
+   If you have not memorized the passphrases to unlock the USB drives
+   for the Secure Viewing Station or the Transfer Device, you may need
+   to access a password manager on your phone or laptop to do so. We
+   recommend switching any required electronic devices into airplane mode,
+   and securely storing any devices you do not need outside the environment
+   in which you access the Secure Viewing Station. This further mitigates
+   the risk of accidentally compromising the air-gap.
 
 Fully mitigating the risks of malware received via SecureDrop is beyond the
 scope of this documentation. If you have questions, you can contact us at
--- a/docs/journalist.rst
+++ b/docs/journalist.rst
@@ -570,11 +570,11 @@ are important steps you can take to protect yourself:
 .. warning::
 
    If you have not memorized the passphrases to unlock the USB drives
-   for the Secure Viewing Station or the Transfer Device, you may need
+   for the *Secure Viewing Station* or the *Transfer Device*, you may need
    to access a password manager on your phone or laptop to do so. We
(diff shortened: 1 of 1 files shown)


eloquence:
The QR code risks are also mentioned further down:

> [QR codes can contain malicious links](https://securedrop.org/news/security-advisory-do-not-scan-qr-codes-submitted-through-securedrop-connected-devices) that your device will automatically visit. In general, be careful when opening any links provided in a SecureDrop submission, as this can leak information to third parties. If you are unsure if a link is safe to click, you should consult internally, or contact Freedom of the Press Foundation for assistance.

Perhaps it makes sense to consolidate these warnings / move the risk section up to be more prominent?

nathandyer:
Thanks @eloquence, I had missed that it was already mentioned on the page. I have consolidated the existing QR warning to include some of the more explicit warnings from my first draft here, and have added a more generic warning about leaving personal electronic devices outside of the area where the _Secure Viewing Station_ is stored. I considered moving the Malware section up to make it more prominent, but I think its position makes sense in the flow of the Journalist guide, so I left it as-is for now. Would be happy to move it further up if folks feel it belongs in a higher spot.

docs/journalist.rst
```diff
@@ -564,6 +567,12 @@ are important steps you can take to protect yourself:
    (potentially including GPS coordinates) to anyone who gains access
    to the file.
 
+.. warning::
+
+   For maximum safety, it is recommended that cell phones and personal 
+   electronic devices be left outside of the secure environment where the
+   *Secure Viewing Station* is used.
```
eloquence, line 574: I think it's a good recommendation, but I will note that to decrypt a file, you'll have to unlock both the Transfer Device and the SVS -- which ideally will have separate passphrases. If folks don't have those passphrases memorized, they'll need to access them from somewhere, and that's most likely another electronic device.

Leaving this open for a bit in case other folks have thoughts on this specific recommendation and how to best word it.

eloquence: @nathandyer I would suggest rewording this as follows:
> If you have not memorized the passphrases to unlock the USB drives for the *Secure Viewing Station* or the *Transfer Device*, you may need to access a password manager on your phone or laptop to do so. We recommend switching any required electronic devices into airplane mode, and securely storing any devices you do *not* need outside the environment in which you access the *Secure Viewing Station*. This further mitigates the risk of accidentally compromising the air-gap.

nathandyer: Thanks @eloquence! I like the suggested rewording, and have gone ahead and committed that change.

eloquence (approved):

PR #239, opened 2021-06-29
Removed LFS requirement from repo and build script

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

* Fixes Issue #234 

    - disabled LFS with `git lfs uninstall`
    - removed lfs tracking for all tracked filetypes with: `git lfs untrack "filetype"`
    - removed and re-added affected files with `git add --renormalize .`
    - committed all changed files
    - reenabled LFS with `git lfs install`
    - verified no files tracked with `git lfs ls-files`
    - removed LFS config in docs/conf.py

## Testing

- [x] Verify that no files are tracked via `git lfs ls-files`
- [x] Check that the RTD build of this branch contains images (spot checks fine)
- [x] CI passing

## Checklist (Optional)

- [x] Doc linting (`make docs-lint`) passed locally

PR #405, opened 2022-12-08
Add NUC12 to the supported hardware list

nathandyer:
## Status

Ready for review 

## Description of Changes

Adds the NUC12 to the list of supported hardware.

Important note: this PR is also the first use of video that I'm aware of. With that being the case, I think it's important we thoroughly test that component, and discuss any structural changes to make sure we're adding in that new functionality in a way that makes sense moving forward.

## Testing

- [ ] Make docs according to the steps in the README
- [ ] Verify that the NUC12 is listed on the Hardware page
- [ ] Verify that the video appears, and that you are able to play it

## Checklist (Optional)


legoktm:
Very briefly, we should enable git-lfs on this repo if we're going to start storing "large" files (and then make sure the deployment end supports it). I think we should also prefer webm files as a more free format (no patent issues).

eloquence:
As a point of information, we previously had LFS enabled for this repo but it caused upstream breakage with ReadTheDocs (see #239 and cross-linked issue) -- it may be more supportable now that we're self-hosting.

nathandyer:
@legoktm I definitely agree about using an open format. I originally was concerned about compatbility (I knew Safari had some issues with webm), but it appears that it's supported in essentially all the browsers as of the latest versions, including Safari. As for specific format, webm does make sense to me for now (I generally think AV1 is a better choice due to increased quality for smaller filesizes, but general support and hardware decoding isn't as widespread yet).

@eloquence I've gone ahead and removed the video portion of this PR for now so we can get this added more quickly. I agree that the deployment story and the content parameters need to be ironed out first before it makes sense to include video. The good news is that we at least know videos are possible, and that it's relatively straightforward to add (at least from the code side).

eloquence (approved):

eloquence:
LGTM, thanks Nathan! Look forward to revisiting the video integration questions.

PR #407, opened 2022-12-12
Add a dark theme

nathandyer:
## Status

Ready for review

## Description of Changes

Adds the `sphinx_rtd_dark_mode` extension, which provides for a toggleable dark theme

## Testing

- [ ] Build according to the instructions in the README
- [ ] Verify that you now see a button in the lower-right to switch between light and dark themes
- [ ] Verify that your theme preference persists after leaving and re-loading the site

## Checklist (Optional)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -149,6 +149,9 @@ html_logo = "../static/i/favicon.png"
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ['_static']
+html_css_files = [
+    'rtd_dark.css',
+]
 
 # Add any extra paths that contain custom files (such as robots.txt or
 # .htaccess) here, relative to this directory. These files are copied
--- a/docs/_static/rtd_dark.css
+++ b/docs/_static/rtd_dark.css
@@ -1,0 +1,116 @@
+/*!
+ * @name          Readthedocs
+ * @namespace     http://userstyles.org
+ * @description	  Styles the documentation pages hosted on Readthedocs.io
+ * @author        Anthony Post
+ * @homepage      https://userstyles.org/styles/142968
+ * @version       0.20170529055029
+ *
+ * Modified by Aloïs Dreyfus: 20200527-1037
+ * Modified by Erik Kalkoken: 20220615
+ * Modified by Nathan Dyer: 20230103
+ */
+
+@media (prefers-color-scheme: dark) {
+
+	a {
+		color: #7dbdff;
+	}
+	a:visited {
+		color: #bf84d8;
+	}
+
+	pre {
+		background-color: #2d2d2d !important;
+	}
+
+	.wy-nav-content {
+		background: #3c3c3c;
+		color: aliceblue;
+	}
+
+	.method dt, .class dt, .data dt, .attribute dt, .function dt,
+	.descclassname, .descname {
+		background-color: #525252 !important;
+		color: white !important;
+	}
+
+	.toc-backref {
+		color: #333333 !important;
+	}
+
+	code.literal {
+		background-color: #2d2d2d !important;
+		border: 1px solid #333333 !important;
+	}
+
+	.wy-nav-content-wrap {
+		background-color: rgba(0, 0, 0, 0.7) !important;
+	}
+
+	.sidebar {
+		background-color: #191919 !important;
+	}
+
+	.sidebar-title {
+		background-color: #2b2b2b !important;
+	}
+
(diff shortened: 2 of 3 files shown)


eloquence:
Looks pretty slick! Looks like it needs a rebase due to the recent requirements update.

The one thing I'll flag is that this theme is pretty low contrast. That could present accessibility issues and would make me a bit nervous about making it the default (we could set `default_dark_mode = False` in `conf.py` to make it optional). Illustrated below.

Curious what @tina-ux and other interested folks think (when they have time to take a look) :) 

## Light Mode
Screenshot from 2022-12-13 17-04-02
## Dark mode
Screenshot from 2022-12-13 17-04-00

nathandyer:
Thanks folks, I definitely agree we want the light theme to be the default. Where I've been testing this out so much on my system, I didn't even realize that it was defaulting to dark! Will add a commit to change that default setting.

gonzalo-bulnes (approved):
- [x] Build according to the instructions in the README
- [x] Verify that you now see a button in the lower-right to switch between light and dark themes
- [x] Verify that your theme preference persists after leaving and re-loading the site

The test plan makes sense and checks out! :white_check_mark: 


I let you merge @nathandyer because I'm not sure if now is a good time or what your plans are :slightly_smiling_face:

nathandyer:
Thanks for the review @gonzalo-bulnes! Not sure why the first commit wasn't signed, but I went ahead and rebased and signed it. Please feel free to re-approve if you have a moment.

I'll speak with @eloquence later today to make sure there aren't any other concerns before we merge.

eloquence:
Thanks @nathandyer for flipping the default. New version looks good to me! Thanks Tina for the thoughtful analysis. On this note:

> I know we don't store cookies and I'm guessing this page is no exception

It looks like this dark mode feature does have a persistent setting that's kept in local storage. Basically, if a reader opts into dark mode, they stay in dark mode until they flip the setting again.

I think that's good for reader convenience, but it does mean that it'll leave a trace behind on the reader's computer. I personally feel that's acceptable since the page would likely also still be in their history and/or cache. @zenmonkeykstop (or others), any concerns from your end on that front?

zenmonkeykstop:
This does represent a new indicator that the user was viewing the SD docs, and while clearing browser cache does remove it, naively just removing the pages from the browser history does not. I'd be inclined to say that we should find a way to address this before merging.

docs/conf.py
```diff
@@ -19,6 +19,7 @@
 # ones.
 extensions = [
     "sphinx.ext.todo",
+    "sphinx_rtd_dark_mode"
```
zenmonkeykstop, line 22: This module stores a pref that persists in local storage - it would be better not to have this present.

zenmonkeykstop (changes requested):

zenmonkeykstop:
On further investigation, one way to achieve this without writing anything would be to provide CSS that overrides styles based on the user's browser-level preference (this is supported in current Firefox and Chrome versions). There's a good example of how to do this here: https://gist.github.com/ErikKalkoken/0baf538c767774bc11c0e8d4028972e6 (tho I kinda prefer the styles in the sphinx theme you used!).

From a security perspective, this is preferable as it doesn't set anything locally. From an accessibility perspective, it follows the user's preferences, which is probably a good practice. The downside is that users can't toggle between themes on the fly, but IMO that's a fair compromise.

nathandyer:
Thanks @zenmonkeykstop! That all makes sense, and I totally agree that at least following the user's preferences (even if there is not a way to manually toggle between the themes) is a good compromise.

Using the CSS media query is the right approach I think. I'll see if I can more-or-less replicate the look and feel of the extension, but directly in CSS. That does further point to something that keeps coming up while looking at these redesigns - it may be in our interests to look at changing the way we theme the docs and diverging away from the RTD theme to something a bit more custom to our needs.

For now, though, I'll see what I can do with the CSS, and will ping folks here when it's ready for another look.

nathandyer:
The latest commit removes the extension, and adds the necessary CSS for this to work natively.

To see this in action, you'll need to be using a modern browser that implements the `prefers-color-scheme` CSS media feature, and have your browser either manually set to dark mode, or have it sync with your OS preferences (and have your OS set to dark mode).

Will also attach a current screenshot:

Screenshot 2023-01-03 at 13-33-54 Welcome to SecureDrop’s documentation! — SecureDrop 2 5 1 documentation

zenmonkeykstop (approved):

zenmonkeykstop:
Looking good! Works for me on FF 102.6.0esr.

PR #408, opened 2022-12-13
Center content when the display is sufficiently large

nathandyer:
## Status

Ready for review

## Description of Changes

This change centers content when the width of the screen is sufficiently large.

To see how this looks (without manually building the docs locally), you can refer to the short screencast below:

[Screencast from 2022-12-13 10-10-21.webm](https://user-images.githubusercontent.com/6131358/207388765-ba5a057d-b10d-4df5-9b15-f77583c3da6b.webm)

This is worth a wider discussion before merging in, but I wanted to make it easy for everyone to visualize the changes, and to have a place to collect feedback and comments.

## Testing

- [ ] Build according to the steps in the README
- [ ] Verify that the content is centered
- [ ] Optionally, test using multiple browser widths to see how it scales and adapts

## Checklist (Optional)

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -33,3 +33,11 @@
     </div>
 </div>
 {% endblock %}
+
+{% block footer %} {{ super() }}
+
+    <style>
+        .wy-nav-content { display: block; margin: 0 auto; }
+    </style>
+
+{% endblock %}


tina-ux:
Heya friend! Making the layout responsive is a great idea, I'm really excited to see this take shape! The video is also a+, thank you @nathandyer.

Before moving forward, there is 1 UX-related aspect that is a blocker for me at the moment: the fixed sidebar in an otherwise responsive and centred layout breaks the user experience in this scenario, especially when we take into consideration the density of our TOC.

I would recommend we implement a layout that centers the TOC with the content, a bit like this template:
https://surjithctly.github.io/documentation-html-template/
However, in order to pull that off, we have to rethink the color scheme and the sidebar design. It's a good opportunity to customize the sphinx template for SD! Unfortunately my schedule and the holidays are a blocker on this one and we will have to tackle it in the new year.
I'm going to assign myself to this issue to keep track of this design task.

Another more minor UX-related element that I would love for us to smooth out more is the transition from one breakpoint to another. It might suffice to play around with css transition animations on the main div along the lines of` transition: all .25s linear; ` etc...
What do you think?

nathandyer:
Thanks @tina-ux!

That all sounds good! I'll make a version of this where the TOC gets centered with the content so we can at least use that as a starting point. Right now we're using the Read the Docs Theme (and are considering the use of the Read the Docs Dark Theme extension), so if we decide to use our own color schemes, we'll have to do a lot of the heavy lifting there on our side rather than using those community-provided packages. That's something we can certainly look into doing, although a change that like will be a lot more involved. Excited to think through the logistics and work towards these bigger changes in the New Year!

As for the smoothness of the transition, the video I recorded has a very low framerate, so it looks quite stuttery. In real life, the transition feels buttery smooth for me. That said, if you think it would be a better experience to set those transitions properties with CSS, I imagine that's something we can do (will have to give it the ol' college try!). 

Thanks as always Tina, this feedback is really helpful!

nathandyer:
After further thought and experimentation, I don't think the benefits here outweigh the downsides, and there are other improvements worth exploring first. Closing this PR for now.

PR #414, opened 2023-01-05
Remove reference to TLS 1.0 and TLS 1.1

loganaden:
https://www.rfc-editor.org/rfc/rfc8996.html has been published since a while.

### DIFF
--- a/docs/deployment/landing_page.rst
+++ b/docs/deployment/landing_page.rst
@@ -361,7 +361,7 @@ to give preference to the most secure cipher suites:
 ::
 
     Header set Strict-Transport-Security "max-age=16070400;"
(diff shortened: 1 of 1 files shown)


nathandyer (changes requested):
Thanks so much for opening this pull request, @loganaden!

The change you made here to the recommended nginx configuration looks correct. I do believe we need to make some changes to the Apache configuration as well to fully remove TLS 1.0 and TLS 1.1.

I believe that can be accomplished by modifying line 364 in your branch to be:

```
SSLProtocol all -SSLv2 -SSLv3 -TLSv1 -TLSv1.1
```

Once that change is in, I'm happy to go ahead and merge this in!

nathandyer (approved):
LGTM! Thanks again!

PR #419, opened 2023-01-17
Update set_up_svs.rst

KwadroNaut:
## Status

It's missing the 'setup persistence step' in the initial svs setup documentation. It does exist at remote.rst Can be solved in many ways to avoid duplication like this.

Ready for review

## Description of Changes

* Description: you need to enable and setup persistence before you can unlock it. This step is m issing.

### DIFF
--- a/docs/set_up_svs.rst
+++ b/docs/set_up_svs.rst
@@ -37,11 +37,10 @@ Station”. If it's not labeled, label it right now, then boot it on the
 *Secure Viewing Station*. After it loads, you should see  the Tails
 `Welcome Screen <https://tails.boum.org/doc/first_steps/welcome_screen/index.en.html>`__.
 
-Enter your passphrase to unlock the persistent storage, then press **Unlock**.
 Before starting Tails, set an administration password for use with this Tails
 session. To do so, click the **+** button under "Additional Settings". Click
 **Administration Password** in the list of settings. Enter the password twice,
-click **Add**, then click **Start Tails**.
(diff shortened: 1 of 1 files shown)


nathandyer:
Thanks so much for this PR, @KwadroNaut!

The lack of instructions for setting up a persistent volume while creating the Secure Viewing Station here is intentional, as that step will already have been completed in the [previous chapter of the install guide.](https://docs.securedrop.org/en/stable/create_usb_boot_drives.html)

We're working on some changes to the documentation to make it clearer that these articles aren't standalone, and are part of an Installation Guide that is meant to be followed in order.

With all that in mind, I would suggest that we go ahead and close this PR for now, unless you have any suggestions for ways we can make this clearer.

KwadroNaut:
Thanks for your feedback. Probably easiest to mention that you need a Tails with persistence, point here, and the same for the remote.rst.

nathandyer:
Thank you @KwadroNaut! I've gone ahead and opened PR #423, which adds an explanation about the expectation of having a Tails drive with persistent storage, and points to the previous step in the installation guide, as you recommended. I'm leaving the more detailed instructions in remote.rst for now, as it's intended to be a standalone article outside of the installation guide.

Please let me know your thoughts, and if the PR I created helps address the need you identified here.

nathandyer:
Hello @KwadroNaut,

I'm going to go ahead and close this PR for now following the merge of PR #423. If there's anything further that you think needs to be done to address this issue, please feel free to reply here to reopen this PR, or file a new PR/issue. Thanks again!

PR #420, opened 2023-01-17
Add a note about using the HWE kernel for NUC12s

nathandyer:
## Status

This adds a note in the NUC12 section of the hardware guide to let folks know to use the HWE kernel for NUC12s.

Just a note that I did not feel it necessary to explicitly recommend the BIOS update here, as it's not strictly required to boot the OS, and it's also [recommended elsewhere in our documentation.](https://docs.securedrop.org/en/stable/update_bios.html)

## Testing
- [ ] Make sure CI is happy
- [ ] Build locally and inspect the Hardware page visually

## Checklist (Optional)

### DIFF
--- a/docs/hardware.rst
+++ b/docs/hardware.rst
@@ -424,6 +424,11 @@ after disconnecting them.
 
 .. _nuc11_recommendation:
 
+Because of the newer hardware and the drivers required, you will need to use a
+newer Linux kernel than the one that ships by default in the version of Ubuntu
+Server we recommend. To do so, select the "Boot and Install with the HWE
+Kernel" option in the boot menu for Ubuntu Server.
+
 Intel 11th-gen NUC
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


docs/hardware.rst
```diff
@@ -424,6 +424,11 @@ after disconnecting them.
 
 .. _nuc11_recommendation:
 
+Because of the newer hardware and the drivers required, you will need to use a
```
zenmonkeykstop, line 427: Should this info be flagged in the install doc as well/instead? Something along the lines of a note just before the Ubuntu install recommending the use of the HWE kernel for NUC12s specifically and newer hardware in general.

nathandyer: That's a good idea. I think it makes sense to flag this specifically in the NUC12 section of the hardware guide since we know that to be needed, but I'll add a general note in the install doc as well.

eloquence (approved):
Previous comment addressed, LGTM.

PR #433, opened 2023-03-03
Fix various links

eloquence:
- Tor GitLab mirror seems to no longer maintained, but they offer a Telegram bot now, so linking to that instead
- Tails has reorganized its persistent storage docs, leading to a bunch of 404s
- ThinkPad redirect seemed to be stalling out, so updating to its redirect target

## Status

Ready for review

## Checklist (Optional)


nathandyer (approved):
LGTM! Thanks!

PR #435, opened 2023-03-08
Update our firewall recommendations

nathandyer:
## Status

Ready for review

## Description of Changes

This expands our firewall documentation to clarify our recommendation to use either pfSense Plus or OPNSense, due to changes in the security updates cadence for pfSense CE

## Checklist (Optional)


eloquence (approved):
Given the current pfSense CE release cadence, this seems indeed a reasonable and necessary update; we can revisit in future if the situation changes.

PR #437, opened 2023-03-14
Fix search

### LINKED ISSUE #436: Search is broken on docs.securedrop.org (and developers.securedrop.org)
## Expected behavior

Entering search term in search bar yields results page

## Actual behavior

Search hangs ("Searching [. .. ...]")

## Steps to reproduce the problem

1. (tested on firefox) visit docs.securedrop.org, enter any term in search bar and hit enter.

## Additional information
workstation.securedrop.org, also a RTD repo, is _not_ facing the same issue. developers.securedrop.org is facing this issue.

legoktm:
## Status

Ready for review

## Description of Changes

* Enable `sphinx_rtd_theme` as an extension, so it can load jQuery
* Upgrade sphinxcontrib-jquery to 4.x so it uses a bundled copy instead of using Google's CDN.

Fixes #436.

## Testing

* Searching (e.g. "monitor") brings up search results
* Check network pane in browser to verify jQuery is loaded from localhost and not a Google CDN

## Release 

* Nope

## Checklist (Optional)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,6 +19,7 @@ import os
 # ones.
 extensions = [
     "sphinx.ext.todo",
+    "sphinx_rtd_theme",
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/poetry.lock
+++ b/poetry.lock
@@ -346,23 +346,6 @@ socks = ["PySocks (>=1.5.6,!=1.5.7)"]
 use-chardet-on-py3 = ["chardet (>=3.0.2,<6)"]
 
 [[package]]
-name = "setuptools"
-version = "67.4.0"
-description = "Easily download, build, install, upgrade, and uninstall Python packages"
-category = "main"
-optional = false
-python-versions = ">=3.7"
-files = [
-    {file = "setuptools-67.4.0-py3-none-any.whl", hash = "sha256:f106dee1b506dee5102cc3f3e9e68137bbad6d47b616be7991714b0c62204251"},
-    {file = "setuptools-67.4.0.tar.gz", hash = "sha256:e5fd0a713141a4a105412233c63dc4e17ba0090c8e8334594ac790ec97792330"},
-]
-
-[package.extras]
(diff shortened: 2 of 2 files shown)


eloquence:
@legoktm Point of process here, should we be specifying the updated dependency in `pyproject.toml` as well? Part of my thinking here is to consider the lockfile as potentially ephemeral -- something that may be regenerated from the input file at a later time -- and if we did this without updating the input file, we'd again break search by doing so.

This also gives us a logical place where we can explain _why_ we're doing this, so we don't lose track of it later.

This is also mirroring what we'd do in a similar situation in the old `.in`/`.txt` process.

I've pushed a commit to a separate branch to explain what I mean: 50e29f65feb387e5994f892cacc293fdee232d37

eloquence:
Nevermind, I think I understand what's happening:
- `sphinxcontrib-jquery` was literally updated today (https://pypi.org/project/sphinxcontrib-jquery/#history)
- the version constraint, for some reason, excludes 3.0.0, but will happily update to 4.x
- if we did rebuild the lockfile from scratch, it would therefore pick up the new version as well, so we don't have to specify it in the input as well.

eloquence (approved):
Per review on https://github.com/freedomofpress/securedrop-dev-docs/pull/41

eloquence:
That makes sense to me, Kunal; this turned out very easy to resolve via `poetry` built-in commands. As long as we are not finding ourselves editing the lockfile directly, no concerns.

cfm:
I second @legoktm's recommendation in <https://github.com/freedomofpress/securedrop-docs/pull/437#issuecomment-1470297418>.  Poetry gives a [high-level rationale](https://github.com/freedomofpress/securedrop-docs/pull/437#issuecomment-1470297418) for committing `poetry.lock`, but I think npm spells out the equivalent (and probably precedential) rationale for the equivalent `package-lock.json` more explicitly and compellingly:

> `package-lock.json` is automatically generated for any operations where npm modifies either the `node_modules` tree, or `package.json`. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.
> 
> This file is intended to be committed into source repositories, and serves various purposes:
> 
> * Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> * Provide a facility for users to "time-travel" to previous states of node_modules without having to commit the directory itself.
> * To facilitate greater visibility of tree changes through readable source control diffs.
> * And optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.  (https://docs.npmjs.com/cli/v6/configuring-npm/package-lock-json)

PR #439, opened 2023-03-29
Include content from SecureDrop.org FAQ in the main documentation

nathandyer:
## Status

Ready for review

## Description of Changes

This change pulls the content from the (somewhat outdated) FAQ at securedrop.org into the main documentation.

While much of the content is the same as what's in the FAQ, it has been re-worked to make sense as part of the existing articles. In one case, I deemed it necessary to create a new documentation page, rather than try to blend that information in with existing ones.

I also have updated some of the information and details, where applicable. While it looks very similar to existing FAQ material, please do take the time to read through it to make sure the parts that have been re-written make sense.

## Testing

* [ ] CI passes
* [ ] Everything builds locally
* [ ] Everything looks OK, including content

## Checklist (Optional)


eloquence:
Hi Nathan, thanks much for working on this. Is the underlying assumption that you'd want to replace https://securedrop.org/faq/ with pointers to the documentation?

nathandyer:
@eloquence Yes, I would be in favor of replacing the FAQ with pointers to the documentation. I might suggest that we add an additional item to the [overview](https://securedrop.org/overview/) giving a fairly broad look at how the system is architected, but otherwise refer folks to the documentation for the more detailed and technical aspects that are currently covered by the FAQ.

eloquence:
(Rebased)

docs/admin/deployment/onion_name.rst
```diff
@@ -0,0 +1,108 @@
...
+
+`sdolvtfhatvsysc6l34d65ymdwxcujausv7k5jk4cy5ttzhjoi6fzvyd.onion`
+
+An onion name for this SecureDrop instance could be:
+
+`nyworld.securedrop.tor.onion`
+
+The general format for a SecureDrop onion name is:
+
+`<organization>.securedrop.tor.onion``
```
eloquence, line 21: Stray backtic. Note that we'll want to use double backtics for preformatted text. I have a slight preference for preformatting over italics here, but either is fine with me.

docs/admin/deployment/onion_name.rst
```diff
@@ -0,0 +1,108 @@
...
+Freedom of the Press Foundation maintains onion names for SecureDrop instances
+which:
+
+* are using v3 onion services
+* are part of the SecureDrop Directory
+
+We will generally approve onion names that meaningfully correspond to your name
+or that of your organization. Please note that, to disambiguate organizations
+in different countries with the same name, we may request the addition of a 
+country code (e.g. `<organization>.<country code>.securedrop.tor.onion`).
```
eloquence, line 54: (Noting that if we switch to preformatted above, we'll want to change it here as well.)

docs/admin/deployment/onion_name.rst
```diff
@@ -0,0 +1,108 @@
...
+No, the onion name is only a human-friendly name for the full-length address.
+The original v3 address can continue to be used like normal, this just gives
+sources an easier to remember option for accessing your SecureDrop.
+
+We recommend that you list both the onion name and the full v3 address on your
+landing page. This allows sources to verify both addresses against the
+information included in our directory, and also provides a fallback should
+the onion name fail to load for any reason.
+
+Please note that as of March 2021, sources need to use the desktop version
```
eloquence, line 77: I think we can remove the "as of March 2021" as we move this into the official docs, and just strive to keep it updated.

docs/index.rst
```diff
@@ -112,6 +112,7 @@ Get Started
 
    admin/deployment/deployment_practices
    admin/deployment/landing_page.rst
+   admin/deployment/onion_name.rst
```
eloquence, line 115: Note that there are two places where we're linking to onion name content on SecureDrop.org:
- in the training schedule (to the blog post)
- in the landing page guidelines (to the FAQ)

I suggest we update both of these to point to the internal docs. In addition, are there other places where we should mention this new section?

docs/admin/installation/installation_overview.rst
```diff
@@ -1,6 +1,29 @@
 Installation Overview
 =====================
 
+Before You Begin
```
eloquence, line 4: I like this placement, but note that we have a "Before You Begin" section later in the installation guide as well, which also speaks to support from FPF.

I would suggest re-titling this section "Installation support" or similar, and consolidating all this information here.

docs/what_is_securedrop.rst
```diff
@@ -3,9 +3,121 @@ What Is SecureDrop?
...
+implementation and branded their version StrongBox in May 2013.
+
+In October 2013, Freedom of the Press Foundation took over management and
+development of the open source project and re-named it SecureDrop. In the
+project's early years at FPF, development was driven by James Dolan and
+`Garrett Robinson <https://github.com/garrettr>`_. 
+Today, SecureDrop is maintained by a small full-time development team at
+FPF and a growing volunteer community.
+
+What Technology Does SecureDrop Use?
```
eloquence, line 43: How about just `Technology` for consistency with the other section headings?

docs/what_is_securedrop.rst
```diff
@@ -3,9 +3,121 @@ What Is SecureDrop?
...
+SecureDrop attempts to completely eliminate third parties from the equation so
+that news organizations can challenge any legal orders before handing over any 
+data.
+
+SecureDrop also substantially limits the metadata trail that may exist from 
+journalist-source communications in the first place. In addition, it attempts
+to provide a safer environment for those communications than regular corporate
+news networks, which may be compromised.
+
+Project History
```
eloquence, line 23: I would suggest we keep the "How it works" pretty high up, maybe just below the "Purpose" section.

eloquence (changes requested):
Hi @nathandyer, thanks for your work on this; I think consolidating this in the docs and ultimately redirecting the FAQ pages here is the way to go.

Left a few comments/suggestions. In addition, I would also suggest that we include the cost information from https://securedrop.org/faq/getting-securedrop-faq/#what-type-of-hardware-is-needed-to-run-securedrop-and-how-much-does-it-cost as this is a common early question folks have.

docs/admin/deployment/onion_name.rst
nathandyer: Thank you! I also prefer the preformatting, I just keep forgetting that RST uses two backticks for it. Fixed :)

docs/index.rst
nathandyer: I updated the links in the two places you described. I can't think of any other places where it might be beneficial to add those links currently, although I can see where it might be useful if we decide to incorporate more of the support documentation in with the main documentation in the future. For now, if you think of any other places where it makes sense, please let me know and I'll get it added.

nathandyer:
Thank you for reviewing, @eloquence! I believe I made all the changes you requested (including adding the section regarding the costs of using SecureDrop). Please let me know if there are any other changes you would like to see.

eloquence (approved):
This looks good to me, Nathan, thanks for your patience. I've pushed one commit, 158d4791bb9cf0e349d1654f50940f49d926acf3, to clarify our advice re: the use of the worksheet, and more visibly put it in its own section. If that looks good to you, feel free to merge!

nathandyer:
Thanks @eloquence, that addition looks great to me! Going to go ahead and merge.

PR #448, opened 2023-05-22
Document Required SecureBoot Settings for Installation

nathandyer:
## Status

Ready for review

## Description of Changes

This branch includes a new section within the Hardware chapter of the Install Guide that briefly explains what SecureBoot is, and the necessary settings for both the server and the workstations. If SecureBoot is not disabled on the servers prior to attempting a SecureDrop install, it will throw an error in the Ansible playbook before you can continue with the install, which comes as a surprise as the docs don't [yet] advise you to make that change before you make it to that point in the install process.

## Testing
 - [ ] CI Passes
 - [ ] Builds locally

## Checklist (Optional)


zenmonkeykstop (approved):
LGTM based on visual review.

PR #469, opened 2023-06-15
Add documentation related to the SecureDrop Menu

nathandyer:
## Status

Ready for review

This PR is ready for wider review, although it should wait to be merged alongside the 2.6.0 release. Keeping this in draft mode for now because of that requirement.

## Description of Changes

This PR documents the new SecureDrop Menu, and refers to it throughout the documentation at the appropriate times when someone might need to make use of its features.

Given the number of times it is referenced, I thought it made sense to add it to the glossary. That said, we can choose to document this outside of the glossary if that is preferred.

It's also worth noting, I included an additional image file that is not explicitly used within the documentation itself (showing a blank Tails desktop with the icons and the extension active). I can see a future time where this might be handy, so I kept it in the folder structure; if it is better to omit, I'm happy to take it out.

## Testing
* [x] CI passes
* [ ] Visual inspection

## Release 

This will need to be released alongside the 2.6.0 SecureDrop release.

## Checklist (Optional)


eloquence (approved):
Also LGTM. `securedrop_menu_desktop.png` appears to not be in use, perhaps it should be dropped?

PR #474, opened 2023-07-18
Merge the existing Support Docs (Redmine Docs) into the SecureDrop Documentation

### LINKED ISSUE #471: Merge existing content from redmine_docs into securedrop-docs
- Merge existing git history from redmine_docs into securedrop-docs (conversation with @legoktm to come)
- update nav etc to reflect new content location

nathandyer:
## Status

Ready for Review

## Description of Changes

This PR merges the git history from the Redmine docs repository and merges the content from the previous repository into the "Getting Support" article that is part of the main SecureDrop documentation.

* Resolves #471 

## Testing
- [x] Git history is in tact
- [ ] Content in the "Getting Support" page matches the relevant information from the old Redmine docs portal
- [x] CI passes
- [x] Visual inspection

## Release 

Prior to merging, we'll want to make sure that there are redirects in place from the old Redmine docs website pointing to this support article.

Once it is merged, we'll need to tag a stable release ASAP.

## Checklist (Optional)


eloquence:
Nathan, just a note that the most recent commits are unsigned.

eloquence:
Overall I think this is coming together nicely. I didn't spot an issue with the commit history, but Kunal's trained eyes will be definitely helpful there.

I'm assuming this is still WIP since it's marked as such; let me know when it's ready for formal review. One suggestion: With all this new content, I think we should add a short intro (spelling out the two main support paths and linking to the appropriate subsections).

nathandyer:
Thank you @eloquence! The commits should be signed and displaying properly here, now. I also added a brief introduction and linked out to some of the sections below, as you recommended.

I think this is officially ready for review whenever someone has time to look at it. Thanks again!

docs/getting_support.rst
```diff
@@ -3,47 +3,461 @@
...
+.. |PGPMenu| image:: images/support/pgp_menu.png
+.. |PGPUpload| image:: images/support/pgp_upload.png
+.. |PGPSave| image:: images/support/pgp_save.png
+
+
+.. _community_support:
+
 Community Based Support
------------------------
+^^^^^^^^^^^^^^^^^^^^^^^
```
eloquence, line 294: This feels a bit sandwiched between Redmine stuff right now, I would suggest moving it to the end or the beginning of the chapter.

docs/getting_support.rst
```diff
@@ -3,47 +3,461 @@
...
+SecureDrop community using the
+`SecureDrop Gitter channel <https://gitter.im/freedomofpress/securedrop>`_ .
 
 .. warning::
 
    Remember that both the SecureDrop forum and the Gitter channel are
    public. **Do not post any sensitive information through public channels.**
+
+Troubleshooting
+^^^^^^^^^^^^^^^
```
eloquence, line 310: I think we'll need to tweak organization and section title here. What are we troubleshooting - Redmine access itself, or SecureDrop? If it's the latter, some of the sub-sections don't really fit. Given that this is a second-level heading, I could see folks getting confused -- "I'm trying to troubleshoot 2FA on SecureDrop, but this seems to be something else entirely"

The old troubleshooting section was always messy and will ultimately benefit from some cleanup. For now, I would suggest:
- Put critical onboarding and usage information in the first chapter (i.e. creating issues, setting up 2FA)
- Put support portal tips and troubleshooting information in the second chapter, under a title like "Support Portal Tips and Troubleshooting" or similar, that leaves now doubt about its purpose

docs/getting_support.rst
nathandyer: Agreed - I decided to move it to the end of the chapter, to encourage folks to use the support portal over the community support (when it makes sense to do so).

docs/getting_support.rst
nathandyer: Thanks! I ended up restructuring this quite a lot, and in the end I realized the information in the current "Troubleshooting" section actually belonged within the main headings rather than a catch-all "Troubleshooting" section. As a result, as of the most recent commit, this section is now omitted entirely. Please let me know if you like this new structure better, and if you think it's necessary to have tips/troubleshooting section here at all.

docs/getting_support.rst
```diff
@@ -3,45 +3,435 @@
...
+Description fields. All other fields can be left blank, although you can
+assign a priority if you want. When we see the issue, we will respond
+and assign it to the appropriate person. If we're waiting for a reply
+from someone in your organization, we may assign it back to you.
+
+|NewIssue|
+
+.. |NewIssue| image:: images/support/new_issue.png
+
+Enabling two-factor authentication
```
eloquence, line 179: I would suggest we make this part of the "creating an account" portion, prior to them testing out ticket submission and such -- this helps establish that 2FA is part of the account setup process, and not something to maybe pick up later.

eloquence:
One structural suggestion, otherwise LGTM pending Kunal's sign-off on the history merge :)

docs/getting_support.rst
```diff
@@ -3,45 +3,435 @@
...
+1. Provide **your public key** to the support server
+2. Import the **server's public key** into your local GPG keyring.
+
+The following documentation explains how to perform these steps to get
+the encrypted email workflow working. This documentation assumes you are
+familiar with PGP/GPG. We encourage you to use your preferred GPG key
+management tool and email client.
+
+If you do not know how to do things like import and export GPG keys,
+the encrypted email workflow is not for you. We recommend you stick to
```
legoktm, line 309: The advice makes sense (don't bother unless you already know PGP), but as written now it feels unnecessarily personal?

"If you are not already familiar with using PGP/GPG keys,we would recommend using the web-based workflow instead of setting up the encrypted email workflow." or something.

docs/getting_support.rst
```diff
@@ -3,45 +3,435 @@
...
+Our server is configured to reject emails that do not have a valid
+signature. We cannot reject unencrypted emails due to a limitation in
+the Redmine-OpenPGP plugin we are using, but we are logging them and
+will harangue you if you send us unencrypted email.
+
+PGP/MIME vs. inline PGP
 -----------------------
+
+When you're replying to Redmine via email, you **must use PGP/MIME**
+instead of inline PGP. The default settings for the Enigmail plugin in
```
legoktm, line 378: Enigmail is dead, "The default settings for Thunderbird should work."

docs/getting_support.rst
```diff
@@ -3,45 +3,435 @@
...
+Thunderbird should work.
+
+Check your mail client and GPG integration tool's settings to ensure you
+are using PGP/MIME. One notable exception to this is Mailvelope, which
+cannot send PGP/MIME (you shouldn't be using Mailvelope anyway).
```
legoktm, line 383: I think we should either have an explicit "don't use Mailvelope" or remove this aside, it doesn't seem helpful as is.

docs/getting_support.rst
```diff
@@ -3,45 +3,435 @@
...
+Per-recipient Rules
+-------------------
+
+You may not want to use these settings (PGP/MIME, always sign) for all
+of your email. If not, most email clients support the concept of
+"Per-recipient rules", which allow you to configure specific settings on
+a per-recipient basis.
+
+Below is an example of how to configure a per-recipient rule for
+support@freedom.press in Thunderbird/Enigmail. If you are not using
```
legoktm, line 394: Just Thunderbird

docs/getting_support.rst
```diff
@@ -3,45 +3,435 @@
...
+Thunderbird/Enigmail, consult your mail client's documentation.
+
+Thunderbird
+-----------
+
+To check your global settings, go to Edit > Account Settings in
+Thunderbird. Navigate to the "OpenPGP Security" tab of the email account
```
legoktm, line 401: The menu item is now called "End-to-End Encryption".

AFAICT there's no longer a PGP/MIME option, I think it's the only thing that's supported?

And...I can't figure out how to set a per-recipient rule. If I compose an email to support@ it does prompt me that encryption is possible, but I haven't yet found an option to enforce it.

legoktm (approved):
Git history merge looks good! I did a quick pass on the content, but I assume we'll fix those in follow-ups since it's not related to the merge itself.

docs/getting_support.rst
nathandyer: Given these changes to Thunderbird (and the fact that the options we're flagging are either removed/non-existent), and that we don't provide these details for any other email clients, I'm going to suggest that we remove this Thunderbird-specific section entirely.

nathandyer:
Thanks for the thorough review, @legoktm! I just pushed a revision that I believe addresses your suggestions. I'm going to leave some of the threads open for now in case we want additional discussion, but if my changes look OK to you, please feel free to mark as resolved.

docs/getting_support.rst
legoktm: I think it's probably worth keeping the warning in some form, "Note that Mailvelope does not support PGP/Mime" (AFAICT it still doesn't: https://github.com/mailvelope/mailvelope/issues/41), I just thought the admonishment "you shouldn't be using it anyways" was out of place.

docs/getting_support.rst
legoktm: Yeah I think that's reasonable. If anything, we could just link to https://support.mozilla.org/en-US/kb/openpgp-thunderbird-howto-and-faq and leave it at that.

Now that we're discussing it, I am curious 1) how many people are using PGP email for redmine and 2) what email clients they're using. The former seems like something we could get out of mail server logs or the redmine DB (checking if people have a key set), the latter would require logging incoming mail before redmine processed it? AFAICT Thunderbird sends the `User-Agent` field in cleartext, idk about other clients. Maybe not worth the effort, but I'm at least curious :)

legoktm:
Besides https://github.com/freedomofpress/securedrop-docs/pull/474#discussion_r1271060407, LGTM :)

legoktm:
The only thing I see missing from the existing redmine_docs is this note at the beginning:
> The [support site](https://support.freedom.press/) is also available as a Tor Onion Service at http://sup6h5iyiyenvjkfxbgrjynm5wsgijjoatvnvdgyyi7je3xqm4kh6uqd.onion/.

Was that an intentional omission? In any case it definitely doesn't need to be that prominent of a note, just a parenthetical would be fine if we want to mention it at all.

nathandyer:
@legoktm Yes, that was an intentional omission here. Because this is now a chapter deeper down within the Overview section of the documentation, having it mention the onion version felt a little out of place. 

That does make me realize that we don't a comparable note for the SecureDrop docs at all, which I definitely think we should. With that being the case, I might suggest we merge this as-is, and then add a similar note on the `index` page, pointing folks to `http://dftlffjdogaragaxkc6jqxpo77s7rrngimyoq7uuq3clowhmttblcoyd.onion`

legoktm:
The onion link I mentioned goes to the equivalent of redmine, not the docs. :) Let me put in a suggestion to show what I mean...

And +1 to having a note on the docs main page saying they're also available via an onion service.

legoktm (approved):
Awesome work!

PR #476, opened 2023-07-22
set `Request-OTR` header per draft-sahib-httpbis-off-the-record-00

cfm:
## Status

Ready for review, but draft pending discussion of the release considerations outlined below.

## Description of Changes

Closes brave-experiments/request-otr-hackathon#1 by adding `Request-OTR: ?1` to the headers we recommend for SecureDrop landing pages.

## Testing
* [ ] The new header configures correctly on Apache.
* [ ] The new header configures correctly on nginx.

## Release 
We should discuss whether or not we want to make this recommendation before:
1. draft-sahib-httpbis-off-the-record-00 has been standardized; and/or
2. `Request-OTR` is followed by browsers other than Brave.

## Checklist (Optional)


rocodes:
Thanks for bringing this to our attention @cfm and @ShivanKaul :) Really appreiate Brave for considering this threat model (AIUI cases where your adversary has some access to your device but not necessarily to your network traffic patterns, eg in cases of domestic violence).

To quickly recap my understanding of this feature, [the following screen](https://brave.com/static-assets/images/optimized/privacy-updates/images/26-otr-prompt.jpg) would be displayed before a Brave user clicks through to a page that sets this header.

I'm open to a discussion here, but I'm uncertain if this is the right threat model for a SecureDrop landing page for a couple of reasons, which mostly boil down to not wanting to confuse or grant a false sense of security to a potential source. 

The threat model of a SecureDrop _may_ include on-device discovery of browser history, but is generally much broader, and includes advanced adversaries capable of network traffic surveillance, etc. My concern is that many people's mental models of internet communications often conflate local browser settings with communications privacy/anonymity overall (anecdotally, I'm commonly asked "do I really need Tor or a VPN if use incognito mode?," for example), so a message saying "view this off the record?" may be misconstrued to "view this content without leaving a trace anywhere."

The other concern I have is that, since using (or preparing to use) SecureDrop is generally a high-stress situation for potential sources, an additional click-through page might cause more stress or confusion, particularly since it is not uniform across all browsers, so the source experience could vary in a way they may not fully understand.

These are my preliminary hesitations for our specific use-case, but again, I'm very open to other perspectives, and I appreciate the work that has gone into this feature.

lsd-cat:
> To quickly recap my understanding of this feature, [the following screen](https://brave.com/static-assets/images/optimized/privacy-updates/images/26-otr-prompt.jpg) would be displayed before a Brave user clicks through to a page that sets this header.
> 

Thanks for investigating this!

> I'm open to a discussion here, but I'm uncertain if this is the right threat model for a SecureDrop landing page for a couple of reasons, which mostly boil down to not wanting to confuse or grant a false sense of security to a potential source.
> 
> The threat model of a SecureDrop _may_ include on-device discovery of browser history, but is generally much broader, and includes advanced adversaries capable of network traffic surveillance, etc. My concern is that many people's mental models of internet communications often conflate local browser settings with communications privacy/anonymity overall (anecdotally, I'm commonly asked "do I really need Tor or a VPN if use incognito mode?," for example), so a message saying "view this off the record?" may be misconstrued to "view this content without leaving a trace anywhere."
> 
> The other concern I have is that, since using (or preparing to use) SecureDrop is generally a high-stress situation for potential sources, an additional click-through page might cause more stress or confusion, particularly since it is not uniform across all browsers, so the source experience could vary in a way they may not fully understand and that may not make it into newsrooms' instructions for sources.
> 

I tend to agree with your assesment, it seems like this could be a significant UX change and that it would modify both the behavior and the experience just on Brave, potentially resulting in a more confused flow and documentation. Whether or not the `Request-OTR` should happen transparently is not an open debate here, but there would be (non SD related) downsides as well with that behavior. I would also wait until we can understand where this proposed standard is going.

rocodes:
@ShivanKaul We had a quick team discussion about this last week, so based on that and @lsd-cat's comment here I don't we will move forward with this change at the moment. However, we're open to revisiting it at a later date, since most of the discussion/concern is about the UX and not as much about the underlying behaviour change itself. Thank you for the suggestion, and for your work on privacy advocacy - we look forward to staying in touch.

PR #477, opened 2023-07-26
Small quality-of-life docs improvements

nathandyer:
## Status

Ready for review

## Description of Changes

This is a small PR that addresses a few minor papercuts that users have reported brushing up against during installation of SecureDrop, and in using our documentation portal.

There are three specific changes:
* It adds a warning against upgrading the Ubuntu Server installer during the installation
* It makes it clear that the admin username and password set during the server installation must be identical on both the Application and Monitor Servers
* It adds a note to the documentation home page (index) letting users know it's available as a Tor onion service (in case Tor browser doesn't suggest it automatically)

## Testing
* [ ] Visual inspection
* [ ] CI passes

## Checklist (Optional)


docs/admin/installation/servers.rst
```diff
@@ -167,6 +179,11 @@ servers:
 
 Select **Done** and press **Enter** to proceed.
 
+.. warning:: The username and password you choose must be used for both the
```
legoktm, line 182:
```suggestion
.. warning:: The username and password you choose must be the same on both the
```

docs/admin/installation/servers.rst
```diff
@@ -167,6 +179,11 @@ servers:
...
+             *Application Server* and the *Monitor Server*. These values will
```
legoktm, line 183: re: "These values...", the real reason they have to be the same is that we only have one field for password, which is a SecureDrop-limitation, not any technical limitation. Maybe:

"When you deploy SecureDrop from your *Admin Workstation* in a later step, you will only be allowed to enter one password, so it must be identical on both servers."

docs/index.rst
```diff
@@ -14,6 +14,9 @@ This documentation is intended for sources, journalists, and administrators.
 If you would like to contribute to SecureDrop, please see our
 `developer documentation <https://developers.securedrop.org/>`_.
 
+.. note:: This documentation is also available as a Tor Onion Service at
+          http://dftlffjdogaragaxkc6jqxpo77s7rrngimyoq7uuq3clowhmttblcoyd.onion/en/stable/.
```
legoktm, line 18:
```
user@dev ~> curl -I 'https://docs.securedrop.org/en/stable/' | grep onion
onion-location: http://dftlffjdogaragaxkc6jqxpo77s7rrngimyoq7uuq3clowhmttblcoyd.onion/en/stable/
```

✔️ that it matches.

legoktm (commented):
Overall looks good, two wording suggestions.

docs/admin/installation/servers.rst
nathandyer: That's a good point, and I like your suggestion. I just pushed a new version with that wording.

nathandyer:
Thanks @legoktm! Just pushed an updated version to include your suggested changes.

legoktm (approved):

PR #484, opened 2023-08-17
docs: eliminate a round trip from AW2 to AW1 and back

cfm:
## Status
Ready for review

## Description of Changes
By using `ssh-add` before `ssh-copy-id` on the new Admin Workstation 2 (AW2), all steps once Tails is installed on AW2 can take place on AW2 with the existing Admin Workstation 1 (AW1) unlocked, rather than running some steps on AW2 with AW1 unlocked and some steps on AW1 with AW2 unlocked.

This adds a couple of shell commands (a possible source of error) but removes 5 steps and some shuffling of thumb drives (a different possible source of error).

### Assumptions
* If we trust AW2 to have AW1 unlocked, then we trust AW2 to load AW1's SSH private key into memory and use it to connect to the Application and Monitor Servers, in order to upload AW2's own SSH public key.
* In other words, each Admin Workstation has its own keypair to allow revocation of individual keys, not because a given Admin Workstation or its owner cannot be trusted to have access to another Admin Workstation's full keypair (in a procedure in which, in fact, they *do* have such access).

## Testing
- Follow the revised procedure.
- [ ] The critical steps (14)‒(18) succeed.  In particular, in step (16), both the Application and Monitor servers reply (e.g.):
    ```sh-session
    amnesia@amnesia:~/Persistent/securedrop$ ssh-copy-id app
    The authenticity of host 'ueo7o3uieted42jisoysbgipgwosofdr54gammbisqexgkhnhksp5rqd.onion (<no hostip for proxy command>)' can't be established.
    ECDSA key fingerprint is SHA256:GYLZ2+HH6NOsAjLWIOMbiYYR9P0oxrW/cmB0+N9fM4w.
    Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
    /usr/bin/ssh-copy-id: INFO: attempting to log in with the new key(s), to filter out any that are already installed
    /usr/bin/ssh-copy-id: INFO: 1 key(s) remain to be installed -- if you are prompted now it is to install the new keys
    
    Number of key(s) added: 1
    
    Now try logging into the machine, with:   "ssh 'app'"
    and check to make sure that only the key(s) you wanted were added.
    ```

## Release
Non-breaking workflow change; no special considerations.

## Checklist (Optional)


nathandyer:
## Testing
* [x]  The critical steps (14)‒(18) succeed. In particular, in step (16), both the Application and Monitor servers reply

## Other Feedback

I have reviewed the text of this PR, and have stepped through the changes to confirm that all the steps work as described. I'm happy with the current state of it.



I like this approach better, and think it's both less time-consuming and more pleasant to work with. I also think the assumptions @cfm listed are valid/reasonable. That said, I'd like to make sure to at least get @zenmonkeykstop's opinion/approval on this modified approach before I officially merge it in.

zenmonkeykstop:
Looking at this now but will have to reinstall my test instance, so it could be a minute or two :)

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -31,60 +31,56 @@ To set up AW2, follow these steps:
...
-8. Reboot AW2, unlock its persistent volume, and set an admin password on the welcome screen.
-9. Open the file manager (**Applications ▸ Accessories ▸ Files**).
-10. Insert AW1. It should show up in the list of storage devices in the file manager under
+9. Reboot AW2, unlock its persistent volume, and set an admin password on the welcome screen.
+10. Open the file manager (**Applications ▸ Accessories ▸ Files**).
+11. Insert AW1. It should show up in the list of storage devices in the file manager under
     a label like "7.0 GB Encrypted". Click the label and enter the drive
     password when prompted to unlock it.
-11. In a terminal, type the following command:
+12. In a terminal, type the following command:
```
zenmonkeykstop, line 52: This is the original command I think but adding in a progress indicator via `--info=progress2` or similar might be good UX.

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -31,60 +31,56 @@ To set up AW2, follow these steps:
...
-    you will add it to the password manager on AW2.
-
-    (You will need to on-board the new admin's 2FA device to complete this step.
-    If this is not possible yet, you can defer it until later.)
-19. Shut down AW1.
-20. Boot AW2, unlock its persistent volume, and set an admin password
-    on the welcome screen.
-21. Boot into AW2 and run the command ``./securedrop-admin tailsconfig`` in
-    ``~/Persistent/securedrop``.
+14. Run the command ``./securedrop-admin tailsconfig`` in ``~/Persistent/securedrop``.
```
zenmonkeykstop, line 62: This step will fail unless a network connection has been configured for AW2.

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -31,60 +31,56 @@ To set up AW2, follow these steps:
...
 
     This will set up desktop shortcuts and SSH access.
-22. Confirm that you are able to access ``mon`` and ``app`` via SSH (``ssh app`` and ``ssh mon``).
-23. Confirm that you are able to access the *Source Interface* and the *Journalist
+15. Insert AW1. It should show up in the list of storage devices in the file manager under
```
zenmonkeykstop, line 65: AW1 was previously inserted and not removed. so this should be superfluous

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -31,60 +31,56 @@ To set up AW2, follow these steps:
...
+    a label like "7.0 GB Encrypted". Click the label and enter the drive
+    password when prompted to unlock it.
+16. In a terminal, type the following commands to authorize the newly created SSH keypair
+    on your servers:
+    ``ssh-add``
+    ``ssh-add /media/amnesia/TailsData/openssh-client/id_rsa``
+    ``ssh-copy-id app``
+    ``ssh-copy-id mon``
+    ``ssh-add -D``
+17. Confirm that you are able to access ``mon`` and ``app`` via SSH (``ssh app`` and ``ssh mon``).
```
zenmonkeykstop, line 75: I would unmount AW1 before this step just in case ssh-add -D doesn't do the needful. Having the AW1 key completely unavailable would be cool. Also consider making the check something like "ssh app hostname" or even making it a one-liner that runs a similar remote command and true/falses based on the output. No need to drop folks into a shell.

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -31,60 +31,56 @@ To set up AW2, follow these steps:
...
+    If this is not possible yet, you can defer it until later.)
+4. Insert the empty AW2 USB drive.
+5. Launch the Tails installer (**Applications ▸ Tails ▸ Tails Installer**) and install Tails on AW2.
    This will delete all data on the AW2 USB drive.
-5. Shut down AW1.
-6. Boot AW2.
-7. Configure its persistent volume (**Applications ▸ Tails ▸ Configure persistent volume**).
+6. Shut down AW1.
+7. Boot AW2.
+8. Configure its persistent volume (**Applications ▸ Tails ▸ Configure persistent volume**).
```
zenmonkeykstop, line 45: Consider the use of the clone persistent volume option in the tails installer instead - downside is that the entirety of AW1 gets copied, which may not be desirable. Upside is a probable further reduction in complexity.

zenmonkeykstop (changes requested):
Some minor points to address but overall the flow works well.

docs/admin/deployment/onboarding_admins.rst
cfm: Sure, I'll aim for #491 while I'm here.

cfm: Clarified this morning: I'm *not* going to aim for #491 while I'm here, since @nathandyer plans to tackle that.  (@nathandyer, probably you'll want to start from this branch to avoid duplicating or conflicting work, then rebase from `main` once this is merged.)

docs/admin/deployment/onboarding_admins.rst
cfm: Added in 9cd8a8b664c1b2fb2ec1f2d551d9092453946ba2.

docs/admin/deployment/onboarding_admins.rst
cfm: Added as step (10) in 9cd8a8b664c1b2fb2ec1f2d551d9092453946ba2.

docs/admin/deployment/onboarding_admins.rst
cfm: Fixed in 9cd8a8b664c1b2fb2ec1f2d551d9092453946ba2.

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -31,60 +31,63 @@ To set up AW2, follow these steps:
...
+    * ``ssh-copy-id mon``
+    * ``ssh-add -D``
+17. From the file manager (**Applications ▸ Accessories ▸ Files**), eject AW1.
+18. Confirm that you are able to access ``mon`` and ``app`` via SSH.  The
+    following commands should produce the following output::
+
+        amnesia@amnesia:~$ ssh app hostname
+        app
+        amnesia@amnesia:~$ ssh mon hostname
+        mon
```
cfm, line 81: I guess we could do—

```suggestion
        amnesia@amnesia:~$ ssh app hostname && ssh mon hostname
        app
        mon
```

—but I favor keeping it one-to-one.

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -31,60 +31,63 @@ To set up AW2, follow these steps:
...
-22. Confirm that you are able to access ``mon`` and ``app`` via SSH (``ssh app`` and ``ssh mon``).
-23. Confirm that you are able to access the *Source Interface* and the *Journalist
+16. In a terminal, type the following commands to authorize the newly created SSH keypair
+    on your servers:
+    
+    * ``ssh-add``
+    * ``ssh-add /media/amnesia/TailsData/openssh-client/id_rsa``
+    * ``ssh-copy-id app``
+    * ``ssh-copy-id mon``
+    * ``ssh-add -D``
```
cfm, line 73: The alternative here is—

```suggestion
        ssh-add
        ssh-add /media/amnesia/TailsData/openssh-client/id_rsa
        ssh-copy-id app
        ssh-copy-id mon
        ssh-add -D
```

—but I favor reserving the code block for the transcript style we use later in step (18).

docs/admin/deployment/onboarding_admins.rst
cfm: See <https://github.com/freedomofpress/securedrop-docs/pull/484#discussion_r1344877654>.

cfm:
Thanks for your review (and patience), @zenmonkeykstop.  Back to you.

PR #492, opened 2023-08-24
Use the new Tails Persistent Storage copy option where applicable

### LINKED ISSUE #491: Use the new option in the Tails Installer to copy the Persistent Volume where applicable
## Describe the change

The latest update of Tails includes changes to the Tails Installer that allows you to copy the Persistent Volume while cloning the current Tails drive.

For operations where we want to copy the entire Persistent Volume, we should prefer this option over the previous rsync-based methods (and update the documentation accordingly).

nathandyer:
## Status

Ready for review

## Description of Changes

This modifies the workflow in a couple key areas to use the new option in the Tails Installer that can copy an existing drive's Persistent Storage during a cloning operation.

It's worth noting that I elected _not_ to use this tool for the backup procedures, instead preferring the current rsync based approach. This is because the current process allows for multiple workstation backups onto a single drive, which the Tails Installer tool does not allow. For that same reason, I have left the previous steps in place regarding the use of `securedrop-admin backup` and `restore`.

The second commit here changes the workflow for making a remote Secure Viewing Station. It does eliminate the use the "primary" USB drive, opting instead to clone the current SVS entirely (from an airgapped system). With this being a change in procedure with security implications, it may not be one we feel comfortable making; if not, I'm happy to remove that commit.

I will also note that I captured screenshots of the new Tails Installer, but did not feel it added to the documentation, so I elected to leave them out. If we believe it's worth including them, I can do so relatively easily.

* Resolves #491 

## Testing
* [x] CI passes
* [x] Visual review

## Checklist (Optional)


docs/admin/deployment/onboarding_admins.rst
```diff
@@ -31,60 +31,49 @@ To set up AW2, follow these steps:
...
+5. Launch the Tails Cloner (**Applications ▸ Tails ▸ Tails Cloner**).
+   Select the option to **Clone the current Tails.** This will delete all data on the AW2 USB drive.
+6. Check the box marked **Clone the current Persistent Storage.**
+7. Click **Install**.
+8. Choose a unique passphrase for AW2 and record it securely.
+9. Shut down AW1.
+10. Boot AW2, set an Admin password on the welcome screen, unlock the Persistent Storage, and
+    enable all the options in the Persistent Storage settings.
+11. Record the new passphrase for AW2 securely.
 12. Generate a new keypair on AW2 using the following command:
```
legoktm, line 50: Does this overwrite the existing keypair from AW1? It seems like a smoother workflow might be:

1. Move AW1 SSH keypair somewhere else
2. Generate new keypair
3. Add AW1 keypair to ssh-agent, run ssh-copy-id
4. Reset ssh-agent + delete AW1 keypair
5. Verify ssh works with `ssh app`, etc.

What do you think?

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -31,60 +31,49 @@ To set up AW2, follow these steps:
...
+    ``ssh-add``
+    ``ssh-add /media/amnesia/TailsData/openssh-client/id_rsa``
+    ``ssh-copy-id app``
+    ``ssh-copy-id mon``
+    ``ssh-add -D``
+16. Confirm that you are able to access ``mon`` and ``app`` via SSH (``ssh app`` and ``ssh mon``).
+17. Confirm that you are able to access the *Source Interface* and the *Journalist
     Interface* using the desktop shortcuts.
-24. :ref:`Initialize a passphrase database <keepassxc_setup>` on AW2.
+18. :ref:`Initialize a passphrase database <keepassxc_setup>` on AW2.
```
legoktm, line 71: Would this database not have been copied over from AW1? Should this say to delete the other admin's credentials and add the new one's?

legoktm (changes requested):
SVS part looks good, I have questions about the Admin Workstation cloning though.

docs/admin/deployment/onboarding_admins.rst
nathandyer: @legoktm I just stepped through this entire process again from scratch (it had been a while), and confirmed that it does *not* overwrite the existing keypair from AW1. Interestingly, it also remembered the Admin PW, and was essentially a perfect copy of AW1. I booted up AW2 for the first time, unlocked persistent storage, the Admin PW was already enabled, and as soon as I logged in I was able to run `ssh app` and access my server.

Unless there's a reason for us to deliberately delete the keypair and generate a fresh keypair on AW2, I think these steps are (mostly) still correct. I'll tack on a commit to make them completely correct, but let me know if we want to do the SSH keypair regeneration.

cfm: @nathandyer, is it that 

https://github.com/freedomofpress/securedrop-docs/blob/5cd6ae2fe228d53b10ffd91298b43810ea3bcfac/docs/admin/deployment/onboarding_admins.rst?plain=1#L52

doesn't overwrite original keypair 1 from AW1? or is it that in

https://github.com/freedomofpress/securedrop-docs/blob/5cd6ae2fe228d53b10ffd91298b43810ea3bcfac/docs/admin/deployment/onboarding_admins.rst?plain=1#L63-L64

we load first the new keypair 2 on AW2 and then the original keypair 1 from AW1?

The latter was my intention in #484 and is still what I understand the current instructions to do.

docs/admin/deployment/onboarding_admins.rst
nathandyer: That's a good catch, you're exactly right.

docs/admin/deployment/onboarding_admins.rst
cfm, line 1: @nathandyer, with this change, don't we lose the following properties?

- https://github.com/freedomofpress/securedrop-docs/blob/5dae3b7bbbc21452436cbb104d403ecdd20be329/docs/admin/deployment/onboarding_admins.rst?plain=1#L6
- https://github.com/freedomofpress/securedrop-docs/blob/5dae3b7bbbc21452436cbb104d403ecdd20be329/docs/admin/deployment/onboarding_admins.rst?plain=1#L64-L66

nathandyer: Ah yes, we do - I had missed that this was a property we specifically wanted. Will add those steps back in, thanks!

docs/admin/deployment/onboarding_admins.rst
nathandyer: Line 52 does indeed overwrite the original keypair (it prompts if you intend to overwrite, and all you have to do is type Y to proceed).

Your understanding is correct, I'll modify these steps to match that intended behavior.

nathandyer:
@legoktm @cfm Digging up this blast-from-the-past to see if we can get it over the finish line. It has been a while since I looked at this, but it appears that the requested changes have been addressed. Does this just need another pass with a review?

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -31,60 +31,43 @@ To set up AW2, follow these steps:
...
+    If this is not possible yet, you can defer it until later.)
+4. Insert the empty AW2 USB drive.
+5. Launch the Tails Cloner (**Applications ▸ Tails ▸ Tails Cloner**).
+   Select the option to **Clone the current Tails.** This will delete all data on the AW2 USB drive.
+6. Check the box marked **Clone the current Persistent Storage.**
+7. Click **Install**.
+8. Choose a unique passphrase for AW2 and record it securely.
+9. Shut down AW1.
+10. Boot AW2 and unlock the Persistent Storage.
+11. Open the KeePassXC database, remove any unneeded credentials from AW1, and store
```
zenmonkeykstop, line 48: This needs clarification - credentials must be deleted, not edited, and the KeepassXC recycle bin cleared, or they'll still be hanging round.

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -31,60 +31,43 @@ To set up AW2, follow these steps:
...
 
-    This will copy *only* the ``securedrop`` directory from AW1 to AW2.
+    (You will need to on-board the new admin's 2FA device to complete this step.
+    If this is not possible yet, you can defer it until later.)
+4. Insert the empty AW2 USB drive.
+5. Launch the Tails Cloner (**Applications ▸ Tails ▸ Tails Cloner**).
+   Select the option to **Clone the current Tails.** This will delete all data on the AW2 USB drive.
+6. Check the box marked **Clone the current Persistent Storage.**
+7. Click **Install**.
+8. Choose a unique passphrase for AW2 and record it securely.
```
zenmonkeykstop, line 45: This should explicitly say the passphrase is for the Persistent Volume and mention recommendations for complexity.

docs/admin/deployment/remote.rst
```diff
@@ -107,38 +104,23 @@ Creating New SVS USB Drives
...
+   When you see the welcome dialog, unlock the persistent storage
+   volume.
+2. Insert the new (empty) SVS USB.
+3. Launch the Tails Cloner (**Applications ▸ Tails ▸ Tails Cloner**).
+   Select the option to **Clone the current Tails**. 
+   This will delete any data on the new SVS drive.
+4. Check the box marked **Clone the current Persistent Storage**.
+5. Click **Install**.
+6. Choose a strong new passphrase for the new Persistent Storage Volume
+   (a 6-word Diceware passphrase is recommended) and record it securely.
```
zenmonkeykstop, line 117: :+1:

zenmonkeykstop (changes requested):

cfm:
@nathandyer, I'm sorry this has sat neglected.  Before you address @zenmonkeykstop's latest feedback, it looks like there are now conflicts (that may also conflict with that feedback).  Let me know if I can help revise or test this once those are resolved.

nathandyer:
Thanks @zenmonkeykstop for the latest review, and @cfm for flagging the merge conflict. I believe I have fixed the conflict, and addressed the requested changes. One thing to note, in the interim since this was filed, we've already updated the SVS section to point users to the upstream Tails' guide (which I _think_ is also new since I originally filed this PR). I've elected to keep that revision rather than the explicit guide I had started here for that section; that said, because the cloning of Admin Workstations is a bit more involved, I think it makes sense to explicitly guide users through each of the steps. Please let me know what you all think of its current form.

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -32,62 +32,57 @@ To set up AW2, follow these steps:
...
+13. Run the command ``./securedrop-admin tailsconfig`` in ``~/Persistent/securedrop``.
 
     This will set up the *SecureDrop Menu* and SSH access.
-16. In a terminal, type the following commands to authorize the newly created SSH keypair
+14. In a terminal, type the following commands to authorize the newly created SSH keypair
     on your servers:
-    
+
     * ``ssh-add``
     * ``ssh-add /media/amnesia/TailsData/openssh-client/id_rsa``
```
cfm, line 65: If we've booted AW2 in new step (10), then `/media/amnesia/TailsData` is AW1's persistent storage.  Don't we still need old step (12) to mount AW1 and old step (17) to unmount it?

cfm (changes requested):
Thanks for revising this, @nathandyer.  I haven't stepped through it manually again, but in reading through it I have one order-of-operations question (inline).  If my question makes sense, it should be an easy fix; if not, let's chat tomorrow and you can let me know what I'm missing!

docs/admin/deployment/onboarding_admins.rst
nathandyer: @cfm I don't believe so; `/media/amnesia/tailsData` should be AW2's persistent storage, not AW1s. Unlike the original instructions, we're never actually copying directly from AW1 to AW2 (I mean that in the spiritual sense, rather than literal - the contents of that persistent volume are being copied, it's just a more transparent operation during the drive creation process). With the Tails Cloner, AW1's persistent storage gets duplicated _as_ AW2's during the cloning (because of the option checked in new step 6). So when you shut down AW1, boot into AW2, and then unlock the storage, at that point AW1 is completely out of the picture.

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -32,62 +32,62 @@ To set up AW2, follow these steps:
...
-    * ``ssh-add /media/amnesia/TailsData/openssh-client/id_rsa``
-    * ``ssh-copy-id app``
-    * ``ssh-copy-id mon``
-    * ``ssh-add -D``
-17. From the file manager (**Applications ▸ Accessories ▸ Files**), eject AW1.
-18. Confirm that you are able to access ``mon`` and ``app`` via SSH.  The
+14. 
+
+    a. Insert the AW1 drive into the computer, then enter the passphrase for its Persistent
+       Storage Volume.
```
cfm, line 64: I think we may still want the step-by-step instructions from:

https://github.com/freedomofpress/securedrop-docs/blob/fea1d1ecb89e5854965fc7c0c3637f68ef02c2eb/docs/admin/deployment/onboarding_admins.rst?plain=1#L49-L52

cfm (changes requested):
Thanks for conferring about this, @nathandyer.  Last volley, I promise: I think we should keep the step-by-step instructions for handling AW1 in step (14), especially since with the Cloner the user now *won't* have had to do other un/lock operations with the file manager.

docs/admin/deployment/onboarding_admins.rst
```diff
@@ -32,62 +32,62 @@ To set up AW2, follow these steps:
...
+    b. In a terminal, type the following commands to authorize the newly created SSH keypair
+       on your servers:
+
+       * ``ssh-add``
+       * ``ssh-add /media/amnesia/TailsData/openssh-client/id_rsa``
+       * ``ssh-copy-id app``
+       * ``ssh-copy-id mon``
+       * ``ssh-add -D``
+    c. Remove the AW1 drive.
```
cfm, line 73: Likewise here:

https://github.com/freedomofpress/securedrop-docs/blob/fea1d1ecb89e5854965fc7c0c3637f68ef02c2eb/docs/admin/deployment/onboarding_admins.rst?plain=1#L74

nathandyer: Great call, sneaky force-push inbound

cfm (approved):
I stepped through this earlier today for the discussion that's led to 86f5d8ef892cb85045336e2030c5178d48e1680c, and with that change this procedure works as intended.  Thanks a bunch, @nathandyer, not least for your patience in shepherding this!

legoktm (approved):
(per Cory's review; didn't look again)

PR #507, opened 2023-10-08
Document the upgrade flag in the Admin Workstation troubleshooting guide

### LINKED ISSUE #501: Document the upgrade flag in the Admin Workstation troubleshooting guide
## Describe the change
Sometimes, when an update doesn't go according to plan, it's necessary to perform a manual update and clear the flag by running:

```
rm ~/Persistent/.securedrop/securedrop_update.flag
```

We document this in each individual upgrade guide, but it can be difficult to find if you don't know to look there. We should consider mentioning it in the Admin Workstation troubleshooting section.

rkraeher:
## Status

Ready for review

## Description of Changes

* Resolves #501 
 

As agreed in the issue (#501) discussion, this PR adds a new article to the "Admin Guide: Maintenance" section of the docs. Right now the new article is essentially a copy/paste of the [performing a manual update](https://docs.securedrop.org/en/stable/upgrade/2.6.0_to_2.6.1.html#performing-a-manual-update) section, but I have used placeholders for any version specific inputs to make the new article more general. There is also additional summary text introducing the article. Although it repeats existing documentation, the intention is to heighten the visibility of this manual update procedure, as for some users it is otherwise hard to find.

### Changes
* Create new article in Admin Guide: Maintenance section: "Troubleshooting Workstation Updates"
* Edit the version-specific text to be more general
* Add some summary

## Testing

- [ ] CI passes
- [ ] Visual review

## Checklist (Optional)

### DIFF
--- a/docs/admin/maintenance/update_workstations.rst
+++ b/docs/admin/maintenance/update_workstations.rst
@@ -1,0 +1,49 @@
+Troubleshooting Workstation Updates
+===================================
+
+This section includes some general troubleshooting instructions for common workstation
+update issues. For additional, version-specific issues and recommended actions,
+please see the relevant admin upgrade guide.
+
+Performing a manual update
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+Sometimes, when an update doesn't go according to plan, it's necessary to perform a 
+manual update and clear the update flag.
+
+If the graphical updater fails and you want to perform a manual update instead,
+first delete the graphical updater's temporary flag file, if it exists (the
+``.`` before ``securedrop`` is not a typo): ::
+
+  rm ~/Persistent/.securedrop/securedrop_update.flag
+
+This will prevent the graphical updater from attempting to re-apply the failed
+update and has no bearing on future updates. You can now perform a manual
+update by running the following commands: ::
+
+  cd ~/Persistent/securedrop
+  git fetch --tags
+  gpg --keyserver hkps://keys.openpgp.org --recv-key \
+   "2359 E653 8C06 13E6 5295 5E6C 188E DD3B 7B22 E6A3"
+  git tag -v <your_securedrop_version>
+
+The output should include the following two lines: ::
+
+    gpg:                using RSA key 2359E6538C0613E652955E6C188EDD3B7B22E6A3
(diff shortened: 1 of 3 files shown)


docs/admin/maintenance/update_workstations.rst
```diff
@@ -0,0 +1,49 @@
...
+    gpg:                using RSA key 2359E6538C0613E652955E6C188EDD3B7B22E6A3
+    gpg: Good signature from "SecureDrop Release Signing Key <securedrop-release-key-2021@freedom.press>" [unknown]
+
+
+Please verify that each character of the fingerprint above matches what is
+on the screen of your workstation. A warning that the key is not certified
+is normal and expected. If the output includes the lines above, you can check
+out the new release: ::
+
+    git checkout <your_securedrop_version>
```
nathandyer, line 40: Although this documentation is (correctly) written to be generalized, I think we should always explicitly refer to the latest stable SecureDrop release here, rather than use `<your_securedrop_version>`. With SecureDrop, it's important that users always be installing the latest version, and we have tools in place to update the version references inside of the documentation when a new version is released.

With that in mind, I'd recommend that we go ahead and specify version `2.6.0` here, then add this new article to our script that we use to bump the version numbers when a new stable release is out.

Adding that should just be one extra line. That change would be made in the `update_version.sh` script, and [a good example is here on this line](https://github.com/freedomofpress/securedrop-docs/blob/5b083dbe2adb8e8fed3db4a2b8b7b8c0ba807647/update_version.sh#L15) (essentially you would just duplicate that line, and replace the article there with `update_workstations.rst`)

rkraeher: @nathandyer thanks for the feedback. I've made the changes you requested :) I tried to run the script locally just to check it but was getting some errors, possibly related to my shell and grep (I use zsh), but the path I added should be correct

nathandyer: Thanks for making those changes @rkraeher! That path does indeed appear to be correct.

nathandyer (approved):
These changes are great, and this is ready to be merged! Thank you so much @rkraeher, we really appreciate the contribution!

PR #510, opened 2023-10-16
Update python:3.9-slim-bullseye image sha, and switch to nginx:mainline-alpine-slim

mig5:
## Status

Ready for review

## Description of Changes

* Description: 

Works towards: https://github.com/freedomofpress/infrastructure/issues/4407

I noted that we were using nginx:mainline-alpine, which seems to currently lack a fix for the curl vuln - however, mainline-alpine-slim doesn't even *have* curl, plus it is a lot smaller, so we might be able to just switch to it anyway? Win win?

## Testing

* Test that you can build the image (`docker build -t securedrop-docs:imageupdate -f deploy/Dockerfile .)
* Test that you can run the image (`docker run -p 127.0.0.1:5080:5080 securedrop-docs:imageupdate`)
* Test basic functionality at 127.0.0.1:5080 ?


maeve-fpf (approved):
Nice! I did not know there was now a `-slim` version.

PR #559, opened 2024-05-17
Use correct term for ossec key

zenmonkeykstop:
## Status

Ready for review

Small tweak for consistent nomenclature.


eloquence (approved):
LGTM; fixed backtics.

PR #569, opened 2024-06-13
docs: document Tor proof-of-work defense via `securedrop-admin sdconfig`

### LINKED ISSUE #568: document Tor proof-of-work defense
## Describe the change

What documentation should we offer for the Tor proof-of-work defense implemented in freedomofpress/securedrop#7175?
- Just how to enable it
- How to monitor it (non-invasively) via `onionprobe`
- More? something else?

## Additional context

cfm:
## Status

Ready for review

## Description of Changes

Closes #568 by:
1. explaining Tor's proof-of-work defense; and
2. giving instructions on how to enable and disable it.

## Testing

- [x] Visual review.
- [x] Is this a fair summary of our decision to enable this feature in `sdconfig` by default going forward?

    https://github.com/freedomofpress/securedrop-docs/blob/1b9209ec5511de0fb4ac34fbb30531cea8242f96/docs/admin/deployment/tor_pow.rst?plain=1#L13-L15

## Release 

This should be called out prominently in the v2.9.0 release notes.

## Checklist (Optional)

### DIFF
--- a/docs/admin/deployment/tor_pow.rst
+++ b/docs/admin/deployment/tor_pow.rst
@@ -1,0 +1,51 @@
+Tor Proof-of-Work Defense on the *Source Interface*
+===================================================
+
+The SecureDrop *Source Interface* is served as an onion service with an
+``.onion`` URL, requiring Tor Browser to access it over the Tor network.  Tor is
+sometimes targeted for denial-of-service (DoS) attacks that can `slow down the
+Tor network as a whole <https://blog.torproject.org/tor-network-ddos-attack/>`_
+as well as burden individual onion services, including SecureDrops.
+
+Tor now includes a `proof-of-work (PoW) defense
+<https://community.torproject.org/onion-services/ecosystem/technology/pow/>`_
+against denial-of-service attacks that can be turned on for individual onion
+services.  As of SecureDrop 2.9.0, new SecureDrops have this feature enabled by
+default, and we encourage all SecureDrop administrators to turn it on for their
+instances.  While this measure can't speed up the Tor network as a whole if it's
+slow, it can protect your SecureDrop from being attacked specifically; and more
+onion services running with this feature helps improve the resilience of the Tor
+network.
+
+
+Enabling the proof-of-work defense
+----------------------------------
+
+Make sure you have :doc:`installed SecureDrop already
+<../installation/install>`.  Then, on the *Admin Workstation*:
+
+.. code:: sh
+
+  cd ~/Persistent/securedrop
+  ./securedrop-admin sdconfig
+
+The prompts will include::
+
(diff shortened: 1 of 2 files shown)


docs/admin/deployment/tor_pow.rst
```diff
@@ -0,0 +1,51 @@
...
+instances.  While this measure can't speed up the Tor network as a whole if it's
+slow, it can protect your SecureDrop from being attacked specifically; and more
+onion services running with this feature helps improve the resilience of the Tor
+network.
+
+
+Enabling the proof-of-work defense
+----------------------------------
+
+Make sure you have :doc:`installed SecureDrop already
```
nathandyer, line 24: Is is strictly necessary for users to have already installed SecureDrop, since they will be prompted for this option anyway during the initial `sdconfig` run?

nathandyer (commented):
This mostly LGTM, I think the explanation here makes sense and is straightforward as to why an admin might want to consider enabling the option. I left one comment regarding the workflow in regards to whether or not SecureDrop has to already be installed before this option can be enabled.

I'm also wondering if perhaps we should link back to this from this section in the docs: https://docs.securedrop.org/en/stable/admin/installation/install.html#configure-the-installation (We don't explicitly mention other options here, such as SSH-over-Tor, so I'm happy either way we decide here)

docs/admin/deployment/tor_pow.rst
cfm: Good point.  I borrowed this language from here—

https://github.com/freedomofpress/securedrop-docs/blob/a26b3fb0e492048de93ab7bbe50f6efc5c6967fa/docs/admin/deployment/https_source_interface.rst?plain=1#L143

—where I suppose the same objection applies: you _can_ enable this right out of the gate, from your initial `securedrop-admin {sdconfig,install}`.  The difference is that we're now enabling `securedrop_app_pow_on_source_interface` by default, which we are not doing for `securedrop_app_https_on_source_interface`.

Let me know what you think of d33e5cb5a3ed27f5c4ab113926c980e2b50253c9, which tries to clarify this.

cfm:
Thanks, @nathandyer!  I've replied to your review feedback inline.

I don't feel strongly either way about linking to these instructions from <https://github.com/freedomofpress/securedrop-docs/pull/569#discussion_r1638794961>.  We want this to be a transparent improvement to SecureDrop's Tor performance, rather than something administrators should have to think about.  (Though we should absolutely flag it in the release notes as a heads-up.)

docs/admin/deployment/tor_pow.rst
nathandyer: Thanks @cfm - I think the changes you made here are excellent!

nathandyer (approved):
With the most recent commit, everything here LGTM.

PR #597, opened 2024-09-25
Remove uses of the word 'master' and standardize on Template Tails USB

### LINKED ISSUE #90: Standardize naming of "Master" Tails USB aka "Primary" Tails USB
The master/primary Tails USB device, originally used to create SVS, Admin WS and Journo WS via cloning, and subsequently used to perform offline updates of the SVS (and/or back up all the workstations) should probably be referred to by one name throughout our docs, and should maybe even be added to the glossary. 

Working on writing clear docs for the Tails Update process, especially for the SVS (see freedomofpress/securedrop#2244) and looking for a clear and concise way to reference this device, and also noticing that it is referred to as mostly the "master" Tails USB but sometimes the primary Tails USB. 

It's important to clarify that this "primary" device is just a utility and doesn't contain any important data (which isn't intuitive, given the name). If I didn't know better, I would maybe think that the Primary Tails USB was another name for the Admin workstation.  

In particular, this section of this docs is a bit confusing: https://docs.securedrop.org/en/stable/backup_workstations.html?backup-the-workstations
 
Any thoughts?

nathandyer:
## Status

Ready for review

## Description of Changes

This PR removes the word 'master' in all public-facing documentation. In most cases the term wasn't adding clarity and could be omitted entirely; in the one instance where it was descriptive, an even more precise term has been used.

As part of this work, I have standardized on "Template Tails USB" based [on this suggestion](https://github.com/freedomofpress/securedrop-docs/issues/90#issuecomment-757108697) (the other suggestion, "Reference," I thought to be too confusing in context). 

* Resolves #90

## Testing
- [ ] CI is happy
- [ ] Visual review

## Checklist (Optional)

### DIFF
--- a/docs/admin/installation/hardware.rst
+++ b/docs/admin/installation/hardware.rst
@@ -25,7 +25,7 @@ For an installation of SecureDrop, you must acquire:
   are using a desktop for the *Secure Viewing Station*).
 * A dedicated network firewall with at least 4 NICs.
 * At least 3 ethernet cables.
-* Plenty of USB sticks: 1 drive for the master Tails stick, 1 drive for each
+* Plenty of USB sticks: 1 drive for the Template Tails stick, 1 drive for each
   Secure Viewing Station, 1 drive for each *Transfer Device*, 1 drive for each
   *Export Device*, and 1 drive for each admin and journalist.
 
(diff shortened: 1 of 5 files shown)


rocodes (approved):
Thanks @nathandyer : lgtm :)

PR #632, opened 2025-02-06
Add noble migration instructions

legoktm:
## Status

Ready for review

## Description of Changes

* Description: Documentation for the noble migration.

## Testing

*

## Release 

* should be published as part of 2.12.0 changes

## Checklist (Optional)

### DIFF
--- a/docs/admin/maintenance/noble_migration.rst
+++ b/docs/admin/maintenance/noble_migration.rst
@@ -1,0 +1,92 @@
+Ubuntu 24.04 (Noble) migration
+==============================
+
+SecureDrops need to be upgraded to the newer Ubuntu 24.04 (Noble)
+operating system. This process is far simpler than past upgrades
+as it has been fully automated.
+
+Administrators have two options, on the following timeline:
+
+* **semi-automated, through March 21, 2025:** Administrators can manually trigger the upgrade and observe the process.
+* **fully automated, after March 21, 2025:** The SecureDrop team will push an update in mid- to late-March that automatically
+  begins the upgrade process on all servers.
+
+The fully automated upgrade is the simplest option, as it requires no action on your part.
+
+We recommend the semi-automated upgrade for larger instances or if you have a non-standard setup as
+the upgrade will happen whenever you choose it, so you will already be available in case something goes
+wrong during the process.
+
+Preparation
+-----------
+
+Since the end of 2024, all SecureDrops have been checking for any potential issues that need to be resolved
+before the upgrade can happen.
+
+If you are receiving notifications about these issues, they must be resolved before the upgrade can take place.
+
+Please see our :doc:`preparation guide <noble_migration_prep>` for more details.
+
+
+What to know
+------------
+
+SecureDrops are currently running the Ubuntu 20.04 (Focal) operating system that
+will stop receiving security updates in April 2025. All SecureDrops must be upgraded
+by then to ensure you continue receiving security patches.
+
+In the past, Administrators needed to perform a full reinstall of SecureDrop to move over
+to the new version; this is no longer necessary. The SecureDrop team has implemented and tested
+a method that allows for in-place upgrades in an automated fashion. A backup is automatically taken
+before the upgrade begins.
+
+It is our goal that this process requires as little Administrator work as possible.
+
+The upgrade can take up to 30 minutes; your SecureDrop will be inaccessible for that duration. It will
+take place shortly after your selected automated restart time, which you can adjust if desired.
+
+If you have any questions, please reach out to Support.
+
+Semi-automated upgrade
+----------------------
+
+* Ensure your Admin Workstation has been upgraded to SecureDrop 2.12
+* Open a Terminal
+* Run ``cd Persistent/securedrop``
+* Run ``./securedrop-admin backup`` to :doc:`take a backup <backup_and_restore>`
+* Run ``./securedrop-admin noble_migration``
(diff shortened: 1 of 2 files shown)


legoktm:
Still WIP; needs some links and other stuff added. A few placeholders around dates need figuring out. Just wanted to post this so people can see the general direction.

nathandyer:
Thanks @legoktm, this looks great! I think it covers all the questions I would have as an admin. The only thing missing that I'm seeing is a recommendation for folks to take a backup, just to be on the safe side. Is there any reason we wouldn't want to advise that beforehand?

legoktm:
> The only thing missing that I'm seeing is a recommendation for folks to take a backup, just to be on the safe side. Is there any reason we wouldn't want to advise that beforehand?

The upgrade script already takes a backup, except it leaves it on the `app` server. Transferring it to the Admin Workstation could be slow, especially if it's SSH-over-Tor. Of course, people who don't have local access probably benefit the most from having a backup on the AW.

It didn't really hit me until now, but in the manual case, we can just automatically start the backup for them as step 0. And then clean it up once we're done. The server will generate its own backup still, but that's not an issue.

legoktm:
Marking this as ready for review. Key highlights:
* terminology is "semi-automated upgrade" and "fully automated upgrade".
* I've tentatively set March 21, 2025 as the deadline to semi-automatically upgrade. Depending on the final release time, that gives folks 2-3 weeks. I think we can revisit this after the release and extend it if necessary.
* Mentioned that we'll publish technical details in the future - I think we should do it before the release, but this allows the docs PR to land first.

nathandyer:
Hypothetical scenario:

I'm an admin who didn't get a chance to run the semi-automated upgrade prior to March 21.

It is now March 23 and I wish to upgrade, but my servers apparently aren't part of the batch that is receiving the rollout.

Am I safe to run the `./securedrop-admin noble_migration` command? What will happen? When my "real" batch comes around, will it attempt the upgrade again or otherwise cause any breakage?

legoktm:
> Am I safe to run the `./securedrop-admin noble_migration` command? 

Yes.

> What will happen? 

The semi-automated upgrade will run as it normally does.

> When my "real" batch comes around, will it attempt the upgrade again or otherwise cause any breakage?

Nope, once you're on noble, you're all set, the script won't do anything.

----

For technical context, https://github.com/freedomofpress/securedrop/blob/develop/securedrop/debian/config/usr/share/securedrop/noble-upgrade.json is the file that controls which upgrades should happen. The semi-automated upgrade works by overwriting this file to say "upgrade everything now" (https://github.com/freedomofpress/securedrop/blob/35e4b2df2b5c3197c605884e078adc6116cffc69/install_files/ansible-base/roles/noble-migration/tasks/main.yml#L13).

So even in the worst-case scenario race condition in which you server receives receives the fully automated instructions to begin the upgrade, and you manually initiate an upgrade at the same time, it'll just work because they use the same instruction file, and start the same systemd unit. (There is some small gotchas around how the `./securedrop-admin` playbook expects to see exactly 2 reboots, so if the automated upgrade has already started, and it only sees one reboot, the playbook might hang, but the actual server upgrade will complete fine, and a second playbook run will succeed.

Or to look at it another way, the semi-automated way basically just starts the fully automatic upgrade and then waits for it to finish, so they can't really conflict.

nathandyer:
Thanks @legoktm, that's excellent. I wanted to make sure that we didn't need to include any additional language to protect _against_ running that command after the deadline. Since there's no harm in doing so, I think this PR is in good shape.

nathandyer (approved):
This PR LGTM. Leaving open for @cfm to review as well before merging.

docs/admin/maintenance/noble_migration.rst
```diff
@@ -0,0 +1,98 @@
...
+
+We recommend the semi-automated upgrade for larger instances or if you have a non-standard setup as
+the upgrade will happen whenever you choose it, so you will already be available in case something goes
+wrong during the process.
+
+Timeline
+--------
+
+Administrators have until March 21, 2025 to perform a semi-automated upgrade, if you want to.
+After that, the SecureDrop team will push an update that begins fully automated upgrades.
```
cfm, line 24: I found myself going back and forth between the options and the timeline to cross-reference them.  That makes me wonder if they'll be clearer combined rather than separate.  That would also mean only one place to be strict about "can" versus "will" and "before" versus "after" (unless I've misunderstood them myself :-).

What do you two think of this?

```suggestion
Administrators have two options, on the following timeline:

* **semi-automated (through March 21, 2025):** Administrators can manually trigger the upgrade and observe the process.
* **fully automated (after March 21, 2025):** The SecureDrop team will push an update in mid- to late-March that automatically
  begins the upgrade process on all servers.

The fully automated upgrade is the simplest option, as it requires no action on your part.

We recommend the semi-automated upgrade for larger instances or if you have a non-standard setup as
the upgrade will happen whenever you choose it, so you will already be available in case something goes
wrong during the process.
```

cfm (commented):
Looks good!  One non-blocking suggestion inline.

docs/admin/maintenance/noble_migration.rst
legoktm: Thanks, I like it. I made one slight adjustment to remove the parenthesis around the dates since it's key information and not an aside.

cfm (approved):

legoktm:
Going to merge now that Nathan + Cory approved it.

PR #643, opened 2025-03-07
Provide better semiautomated noble migration instructions

legoktm:
## Status

Ready for review

#### Provide better semiautomated noble migration instructions
    
* Tell people upfront the expected time commitment.
* Have people check for updates, as it's straightforward if they're already on the terminal and there's a decent chance we do more point releases.

#### Use "semiautomated" (no hyphen)
    
Per FPF style, and also what we'll be using in all other communications with admins and journalists.

## Testing

* [ ] visual review

## Release 

* for 2.12

## Checklist (Optional)

### DIFF
--- a/docs/admin/maintenance/noble_migration.rst
+++ b/docs/admin/maintenance/noble_migration.rst
@@ -7,13 +7,13 @@ as it has been fully automated.
 
 Administrators have two options, on the following timeline:
 
-* **semi-automated, through March 21, 2025:** Administrators can manually trigger the upgrade and observe the process.
+* **semiautomated, through March 21, 2025:** Administrators can manually trigger the upgrade and observe the process.
 * **fully automated, after March 21, 2025:** The SecureDrop team will push an update in mid- to late-March that automatically
   begins the upgrade process on all servers.
 
 The fully automated upgrade is the simplest option, as it requires no action on your part.
 
-We recommend the semi-automated upgrade for larger instances or if you have a non-standard setup as
+We recommend the semiautomated upgrade for larger instances or if you have a non-standard setup as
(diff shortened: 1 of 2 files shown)


zenmonkeykstop (approved):
LGTM based on visual review

docs/admin/maintenance/noble_migration.rst
```diff
@@ -47,16 +47,19 @@ take place shortly after your selected automated restart time, which you can adj
...
 
-* Ensure your Admin Workstation has been upgraded to SecureDrop 2.12
-* Open a Terminal
-* Run ``cd Persistent/securedrop``
-* Run ``./securedrop-admin backup`` to :doc:`take a backup <backup_and_restore>`
-* Run ``./securedrop-admin noble_migration``
+We recommend setting aside at least an hour for this process.
+
+* Log into your *Admin Workstation* and open a terminal window.
+* Run ``cd Persistent/securedrop``.
```
nathandyer, line 56: Something that didn't occur to me in the previous review - should this path be `~/Persistent/securedrop`? This will almost always work either way, but there's a chance the user could have opened a Terminal and navigated away from home, at which point that wouldn't get them to the right place.

nathandyer (approved):
LGTM, left a note about one possible change but no strong feelings, so happy to merge if we think this is good as-is.

docs/admin/maintenance/noble_migration.rst
legoktm: I think if we just asked them to open a terminal, it's reasonable to assume that they didn't go somewhere else first.

I think we can cater to the more common case that they're doing exactly what we asked, which means slightly less typing for everyone else.

PR #649, opened 2025-03-17
Update HWE recommendations

### LINKED ISSUE #646: Test whether noble needs HWE kernel for NUC 12/13
## Describe the change
We already have guidance that NUC12+ should use the HWE kernel under focal. And if we were staying with focal, I would suggest we make that way more prominent/explicit since most people would be running into it.

But since we're moving to noble, we should double-check whether our HWE advice is correct still - my guess is that 12/13 should be fine using the stock noble kernel.

nathandyer:
## Status

Ready for review

## Description of Changes

This PR removes the HWE recommendation for NUC12 and 13 (no longer needed with Ubuntu Noble), and adds a general note that newer hardware may require use of the HWE.

Follow-up task that closes #646 

## Testing
- [ ] CI passes
- [ ] Visual review

## Release 

Should be added alongside Noble support, if possible

## Checklist (Optional)

### DIFF
--- a/docs/admin/installation/hardware.rst
+++ b/docs/admin/installation/hardware.rst
@@ -439,6 +439,13 @@ to choose from. We currently recommend the 11th through 13th generation NUC mode
     extraneous hardware as physically possible from your servers. This
     could include: speakers, cameras, microphones, fingerprint readers,
     wireless, and Bluetooth cards.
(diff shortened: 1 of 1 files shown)


legoktm (approved):
Thanks!

PR #656, opened 2025-03-31
Recommend viewing journalctl logs in reverse order

legoktm:
## Status

Ready for review

## Description of Changes

Recommend viewing journalctl logs in reverse order so that way the error is at the top.

## Testing

* [ ] visual review
* [x] I already tested that `-r` works as expected

## Release 

* n/a

## Checklist (Optional)

### DIFF
--- a/docs/admin/maintenance/noble_migration.rst
+++ b/docs/admin/maintenance/noble_migration.rst
@@ -106,8 +106,8 @@ Technical details and debugging
 -------------------------------
 
 If something goes wrong, logs can be seen by logging into the servers and
-running ``sudo journalctl -u securedrop-noble-migration-upgrade``. If the error isn't clear,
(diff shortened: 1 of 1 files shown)


nathandyer (approved):
Straightforward and to the point, LGTM.

I think there's a question as to whether reverse chron is the best choice - I like it chronological so the errors are near the bottom, and I don't have to pipe it to `more` or `less` so I can see the relevant parts. But since this is based on feedback, I'm game to try and it and seeing it that makes folks' lives easier.

legoktm:
I agree, I find reverse chronological annoying, but I think telling people to use `-r` is going to be easier than "use the less pager to go to the bottom" or some arbitrary limit. Most likely the error will be on a single line so it shouldn't be too bad, I hope.

PR #659, opened 2025-04-09
Add warning for non-recommended hardware

ChumOfChance:
## Status

Ready for review / Work in progress

## Description of Changes

* Description: 

Adds a warning that suggests performing a semiautomated upgrade before the deadline in the case of using non-recommended hardware for Application and Monitor servers.

## Checklist (Optional)

### DIFF
--- a/docs/admin/maintenance/noble_migration.rst
+++ b/docs/admin/maintenance/noble_migration.rst
@@ -16,7 +16,7 @@ We recommend the semiautomated upgrade for larger instances or if you have a non
 the upgrade will happen whenever you choose it, so you will already be available in case something goes
 wrong during the process.
 
-.. warning:: If you are using `non-recommended hardware <../installation/hardware.html#specific-hardware-recommendations>`_ for your Application and Monitor Servers, we strongly recommend doing a semiautomated upgrade **before** April 15th, 2025. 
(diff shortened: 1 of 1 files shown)


docs/admin/maintenance/noble_migration.rst
```diff
@@ -16,6 +16,8 @@ We recommend the semiautomated upgrade for larger instances or if you have a non
 the upgrade will happen whenever you choose it, so you will already be available in case something goes
 wrong during the process.
 
+.. warning:: If you are using `non-recommended hardware <../installation/hardware.html#specific-hardware-recommendations>`_ for your Application and Monitor Servers, we strongly recommend doing a semiautomated upgrade **before** April 15th, 2025.
```
nathandyer, line 19: Two small nits here:

1) Anytime we use a term that is from the [glossary](https://docs.securedrop.org/en/stable/glossary.html) we make sure to italicize it (*Application* and *Monitor Servers*)

2) Rather than saying "If you are using 'non-recommended hardware <link>`", it is probably better to say "If you are not using 'recommended hardware <link>". This is a very subtle difference, but the former implies we recommend *against* the hardware, whereas the latter implies it just *isn't* recommended.

```suggestion
.. warning:: If you are using not using `recommended hardware <../installation/hardware.html#specific-hardware-recommendations>`_ for your *Application* and *Monitor Servers*, we strongly recommend doing a semiautomated upgrade **before** April 15th, 2025. 
```

nathandyer (changes requested):
Thanks so much @ChumOfChance, this looks great! Two small nits (feel free to accept the suggestion if it works for you), then I'll gladly merge this.

nathandyer (approved):

PR #660, opened 2025-04-09
Install version-pinned Poetry via Pip

eloquence:
This is to ensure that we can stay in sync with the version used by Dependabot

Also bumps Alpine image.

## Status

Ready for review

## Description of Changes

Installs Poetry itself via pip, as opposed to Debian package. Poetry's Debian package tends to get stale quickly (Bookworm has 1.3.2 from ~2022), while Dependabot has been quicker to update, which can lead to lockfile formats falling out of sync.  Pinning to a fixed version number avoids this issue (with the downside that we'll have to manually bump the version in case of security issues in Poetry itself).

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -1,16 +1,34 @@
 FROM debian:bookworm AS sphinx
 
 ARG GIT_BRANCH=main
-RUN apt-get -q update && apt-get -qy upgrade && apt-get -qy install git make latexmk texlive-latex-extra python3-poetry
-COPY ./ .
-RUN poetry install
+RUN apt-get -q update && \
+    apt-get -qy upgrade && \
+    apt-get -qy install \
+      git \
+      make \
+      latexmk \
+      texlive-latex-extra \
+      python3 \
+      python3-venv
+
+# Create a dedicated venv for Poetry and install pinned version compatible with Dependabot
+RUN python3 -m venv /opt/poetry-venv
+RUN /opt/poetry-venv/bin/pip install --upgrade pip && \
+    /opt/poetry-venv/bin/pip install "poetry==2.1.2"
+
+# Make the 'poetry' command available system-wide so Makefile targets work
+RUN ln -s /opt/poetry-venv/bin/poetry /usr/local/bin/poetry
+
+COPY . .
+RUN poetry install --no-root
 RUN deploy/build $GIT_BRANCH
 
-# sha256 as of 2024-06-10
-FROM nginx:mainline-alpine-slim@sha256:244d37691a469d45349d9f29e8b7462d9f510b70c0c93acc5d23ee227070c962
+# sha256 as of 2025-02-14 image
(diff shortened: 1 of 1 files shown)


legoktm:
> Poetry's Debian package tends to get stale quickly (Bookworm has 1.3.2 from ~2022), while Dependabot has been quicker to update, which can lead to lockfile formats falling out of sync.

Is this a problem though? dependabot updates the lockfile to the newest format, but Debian's poetry can still install the old version without issues. trixie has poetry v2, which IMO we can switch to now, and be up to date again, and assume that the poetry devs aren't looking to do another major version bump quickly.

My preference for just getting poetry from Debian is just that I would prefer that we didn't have to care about or manage updating poetry ourselves, in a separate, out of band way that dependabot can't do for us (multiplied by 3 repos).

Correct me if I'm wrong, but I think the migration away from Codefresh will be happening soon-ish, right? Assuming we're moving that into GHA, I would suggest we defer changing how poetry is bootstrapped until then (aside from upgrading to trixie, which should be drop-in), at which point we can probably adopt what Ethan implemented in [actionslib](https://github.com/freedomofpress/actionslib/tree/main/act/poetry).

eloquence:
I do think you're correct that we can merge Dependabot updates; it outputs a compatibility warning but installs successfully. I haven't merged them yet out of an abundance of caution but I think we can safely proceed even on Bookworm, for now. And CI should tell us about any compat issues for future Dependabot PRs.

I hadn't seen the `actionslib` implementation yet; that would indeed be very clean, and yes, infra is working on the Codefresh phase-out this quarter.

Let me check in with @enpaul here; given that this is one of our simpler deploy pipelines, maybe it can be an early test case.

deploy/Dockerfile
```diff
@@ -1,16 +1,34 @@
...
+      make \
+      latexmk \
+      texlive-latex-extra \
+      python3 \
+      python3-venv
+
+# Create a dedicated venv for Poetry and install pinned version compatible with Dependabot
+RUN python3 -m venv /opt/poetry-venv
+RUN /opt/poetry-venv/bin/pip install --upgrade pip && \
+    /opt/poetry-venv/bin/pip install "poetry==2.1.2"
```
enpaul, line 17: I think ultimately this is a security downgrade compared to installing via apt (though I agree that installing from pypi is better overall). Installing this way doesn't guarantee that we get the same transient dependency versions in every build since pip doesn't make any guarantees about dependency resolution reproducibility.

In general, this is what a lockfile is supposed to solve, and a solution like the one implemented [here]() in actionslib or [here]() in fpf-www-projects would add that. Another option would be to treat another requirements file (for example the one in actionslib) as the "canonical" requirements file. Instead of re-implementing the lockfile+install logic, you could just use `ADD https://raw.githubusercontent.com/path/to/requirements.txt /opt/poetry-venv/requirements.txt` and then do a `pip install -r /opt/poetry-venv/requirements.txt` from there.

eloquence:
Thank you both. Closing in favor of #663, and then we migrate to `actionslib` later as part of the Codefresh phase-out.

PR #663, opened 2025-04-17
Move to Debian Trixie

### LINKED ISSUE #580: Set `package-mode = False` once we use a compatible version of Poetry
On recent versions of Poetry, `poetry install` warns:

> Warning: The current project could not be installed: No file/folder found for package securedrop-docs
> If you do not want to install the current project use --no-root.
> If you want to use Poetry only for dependency management but not for packaging, you can disable package mode by setting package-mode = false in your pyproject.toml file.
> In a future version of Poetry this warning will become an error!

We cannot set `package-mode = false` yet because we use the version of Poetry included in Debian Bookworm ([1.3.2](https://packages.debian.org/bookworm/python3-poetry) as of this writing), and the setting was only added in Poetry 1.8.0, [released](https://python-poetry.org/blog/announcing-poetry-1.8.0/#better-support-for-non-package-projects) in February 2024.

Unless we install Poetry from a non-Debian source, we'll likely have to wait until Trixie to make this change.

eloquence:
As of this writing, it is still pre-release, but stable enough for documentation repos. This also gives us a more recent version of Poetry, which in turn allows us to disable package-mode, and mitigats lockfile format drift with Dependabot.

## Status

Ready for review

## Description of Changes

Resolves #580

## Checklist (Optional)

- [x] Have also built Docker image locally

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     # We use a standard Debian image to mirror a typical developer environment.
     # This should be updated whenever a new Debian stable version is available.
-    container: debian:bookworm
+    container: debian:trixie
     steps:
       - uses: actions/checkout@v4
       - name: Install dependencies
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     # We use a standard Debian image to mirror a typical developer environment.
     # This should be updated whenever a new Debian stable version is available.
-    container: debian:bookworm
+    container: debian:trixie
     steps:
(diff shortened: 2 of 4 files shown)


legoktm (changes requested):
The linkcheck CI job also needs to be updated.

Otherwise LGTM. Note that this will also move the dev-docs and workstation-docs CI over to trixie because those repos just use these CI pipelines. I don't know if that will cause issues given the new poetry version, or if we need to migrate them first?

(IIRC last time we did this juggle of updating those CI pipelines to use bookworm and CI via a `@bookworm` branch, then merging this one, and then changing those back to `@main`. Would be nice to avoid but also not a big deal, it's only once every 2 years)

legoktm:
LGTM, will merge this after the 2.12.2 release.

legoktm (approved):

PR #700, opened 2025-07-10
update ssh restart command for Noble

### LINKED ISSUE #698: The command to restart SSHD should be "sudo service ssh restart" rather than "sudo service sshd restart"
In [this document](https://docs.securedrop.org/en/stable/admin/maintenance/rebuild_admin.html#configuring-access-for-an-ssh-over-tor-instance) we advise restarting SSHD via `sudo service sshd restart`, but that service does not exist.

Running `sudo service ssh restart`, however, does the trick.

We should update this in the docs.

ChumOfChance:
## Status

Ready for review 

## Description of Changes

* Description: 
Fixes the command to restart `ssh` in accordance with Ubuntu 24.04's change in how sshd is managed by systemd. The `service` command is also replaced with `systemctl`.

* Resolves #698
 

## Checklist (Optional)

### DIFF
--- a/docs/admin/maintenance/rebuild_admin.rst
+++ b/docs/admin/maintenance/rebuild_admin.rst
@@ -165,7 +165,7 @@ and deleting the line:
 
   PasswordAuthentication no
 
-Restart ``sshd`` using the command ``sudo service sshd restart``.
+Restart ``sshd`` using the command ``sudo systemctl restart ssh``.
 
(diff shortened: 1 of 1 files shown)


nathandyer:
Closing this PR in favor of https://github.com/freedomofpress/securedrop-docs/pull/701 which includes this fix within its scope

PR #712, opened 2025-08-21
Metadata cleaner update

### LINKED ISSUE #710: Update metadata cleaning docs
The docs in https://docs.securedrop.org/en/stable/journalist/documents.html are fairly outdated:

- Still reference changes in Tails 4
- Refer to a context menu in Nautilus that no longer exists ("Open With -> Metadata cleaner" works as an alternative)
- Probably other outdated bits

They also suggest that the metadata cleaning tools make a copy - this is not true for the current Metadata Cleaner, which warns that "Once the files are cleaned, there's no going back".

ChumOfChance:
Fixes #710 

Updates to Remove Metadata section which was out of date regarding packaged GUI options for metadata removal. 
Note I have used screenshots from Tails 7.0~rc1 to avoid the need change them again in a few weeks.


docs/journalist/documents.rst
```diff
@@ -114,39 +114,30 @@ documents, pdf, tar, tar.bz2, tar.gz, zip, mp3, mp2, mp1, mpa, ogg,
...
-line).
+You can use MAT2 via the *Metadata Cleaner* application which is found in the **Accessories** section of the **Apps** menu.
 
-You can use MAT2 via the *Files* application by browsing to **Places ▸
-(Your file's location)** and right-clicking on your file. In the context
-menu, select **Remove metadata**.
+|Metadata Cleaner menu|
 
-|mat2 context menu|
+You can load individual files or folders into Metadata Cleaner by clicking Add Files or Add Folders. Metadata Cleaner gives you a count of how many metadata parameters are associated with each file and lets you view the metadata.
```
nathandyer, line 121:
```suggestion
You can load individual files or folders into *Metadata Cleaner* by clicking **Add Files** or **Add Folders**. *Metadata Cleaner* gives you a count of how many metadata parameters are associated with each file and lets you view the metadata.
```

Button names should be bolded to make it clear that these are literal choices within the application. Also, it's worth consistently italizing the application name.

docs/journalist/documents.rst
```diff
@@ -114,39 +114,30 @@ documents, pdf, tar, tar.bz2, tar.gz, zip, mp3, mp2, mp1, mpa, ogg,
...
 
-Note that this does not alter the original file--it creates a clean copy.
+|Metadata Cleaner loaded|
 
-|mat2 cleaned|
+You can select Clean to remove the metadata from the files. Note that this process deletes the original files, leaving only the cleaned versions.
```
nathandyer, line 125:
```suggestion
You can select **Clean** to remove the metadata from the files. Note that this process deletes the original files, leaving only the cleaned versions.
```

nathandyer (changes requested):
This is looking excellent! Two small nits, but otherwise in good shape to be merged.

PR #713, opened 2025-08-21
Update Ubuntu version to 24.04.3

### LINKED ISSUE #711: Ubuntu Server download links are currently broken
It appears that Ubuntu 24.04.2 is no longer available from Ubuntu, and has instead been replaced with the .3 release. 

The download links currently return 404s.

We should update this ASAP so that the links are accurate and functional

nathandyer:
Fixes #711

## Checklist

This change accounts for:


nathandyer:
Thanks @ChumOfChance, definitely worth bumping those as well. Great catch!

Force-pushed a commit which should include that now.

PR #716, opened 2025-08-22
Update the hardware guide, with a focus on updating SVS recommendations

### LINKED ISSUE #479: Update hardware recommendations for SVS
the SVS requires a fully airgapped machine, preferably a laptop to avoid the use of peripherals. Laptops with removable networking cards are becoming rarer, so we should identify viable candidates and update our recommendations.

nathandyer:
Fixes #479

This PR:
* Adds more modern recommendations for specific Secure Viewing Station models
* Explicitly warns against M-series Macs, and gently discourages Macs in general
* No longer encourages users to buy an external hard drive for decrypted submissions on the SVS
* Recommends larger USB drives for the SVS, to avoid the need for said external hard drives

## Checklist

This change accounts for:

### DIFF
--- a/docs/admin/installation/hardware.rst
+++ b/docs/admin/installation/hardware.rst
@@ -35,7 +35,7 @@ Additionally, you may want to consider the following purchases:
 
 * a printer without wireless network support, to use in combination with the
   *Secure Viewing Station*.
-* an external hard drive to expand the storage capacity of the
+* an external storage drive to expand the storage capacity of the
   *Secure Viewing Station*.
 * an external hard drive for server backups.
 * a USB drive to store :ref:`backups of your Tails workstation drives <backup_workstations>`.
@@ -76,6 +76,11 @@ and recycled ThinkPad laptops work well for the
 
 Please note that very old laptops or desktops may not work for the workstations.
 Since the release of Tails 3.0, 32-bit computers are no longer supported.
+
+Additionally, we recommend against re-purposing Apple Macintosh
+laptops and desktops. Recent Apple computers, which have M-series CPUs, are
+completely unsupported by Tails and will not work. Older computers with Intel
+CPUs may work, but are likely to experience compatibility issues.
 
 If you choose to use recycled hardware, you should of course
 consider whether or not it is trustworthy; making that
@@ -160,8 +165,11 @@ SecureDrop and to process submissions using the air-gapped workflow.
 1 physical computer used as an air-gap to decrypt and view submissions retrieved
 from the *Application Server*.
 
-The chosen hardware should be solely used for this purpose and should have any
-wireless networking hardware removed before use.
+The chosen hardware should be solely used for this purpose. This system must have
+a removable wireless card, which will need to be completely removed before use. It
+must also have a removable storage device, such as a solid state drive (SSD) or a
+hard disk drive (HDD), which likewise needs to be removed prior to using the system.
+You will only ever boot directly to the *Secure Viewing Station* USB drive.
 
 Admin/Journalist Workstation(s)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


eloquence:
@nathandyer This mostly LGTM. Regarding no longer recommending an external drive for offline storage, what's the motivation for that? I don't see it mentioned in the original issue. USB drives aren't really designed for long-term storage and generally less reliable than SSDs and HDDs, plus I worry that even the 128GB increase of the USB drive size won't be enough for orgs that routinely deal with large submissions.

docs/admin/installation/hardware.rst
```diff
@@ -627,13 +625,21 @@ not need a hard drive or network device; in fact, we recommend removing
...
 removable.
 
+If you are purchasing a new laptop for use as a *Secure Viewing Station*,
+you will want to ensure you are purchasing a model with a removable
+wireless card and removable storage. We have tested the following laptops
+with removable components, and can confirm compatibility within Tails:
+
+* Framework 13 laptop (Intel)
+* Lenovo Thinkpad X1 Carbon (10th generation)
+* Lenovo Thinkpad T14 (2nd generation)
```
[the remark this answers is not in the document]
eloquence: That makes sense! How about wording like this for the X1:

> For other models, we recommend that you check the service manual for the specific generation/model of laptop. For example, according to the service manual for the X1 carbon, it uses a standard removable M.2-based wireless card.

(Note that the link above points to the T14/P14.)

eloquence:
>  I was thinking that because backups are stored on the Admin Workstation first, maintaining an additional large drive was not a tremendous value, especially when considering it comes with its own risks as an additional sensitive asset that must be protected. That said, if organizations are storing multiple backups for historical purposes, then that recommendation does make sense.

Our backup docs advise against treating the servers as a long-term archive due to Tor slowness, and instead frequently deleting submissions. Of course, there are other security reasons not to do so as well (making it harder for adversaries to acquire vast copies of content that could be decrypted in future; avoiding accumulation of metadata). 

The flip side of that recommendation is that for news organizations that follow it, `./securedrop-admin backup` at any given time will create a snapshot of recently active sources, not an archive of all sources. 

Re-constructing from a series of  server backups a complete archive of valuable submissions requires keeping deletion/backup operations in sync, which some organizations may do, but is probably not a safe assumption.

In addition, it's only on the SVS that spam can really be sorted out from ham, so the SVS may be a better "pristine" source for producing an archive that is the stuff the organization wants to keep: it may be both _more_ complete in covering all useful submissions, and _less_ complete in not including unwanted spam.

> I do think that given the plummeting cost per GB of solid state storage, keeping the upgraded recommendation for flash drive sizes is still worthwhile.

I agree.

> I'm also iffy on whether to explicitly recommend a hard drive over an SSD for backup storage. 

I would distinguish between extending hot storage capacity (for which SSD is fine) and cold storage backups. Archivists generally only recommend SSD for hot storage, but even a single HDD isn't really sufficient for long-term preservation.  As long as we don't recommend SSD for cold storage, recommending it simply to extend the capacity of the SVS seems reasonable to me. Any additional secure backups and fixity checks are probably beyond the scope of what we can cover.

eloquence:
Looks good! Two minor nits:
- It looks like the double parenthesis fix `((solid state drive (SSD) or hard disk 
drive (HDD))` from earlier got accidentally clobbered.
- `carbon` -> `Carbon`

nathandyer:
Thanks @eloquence, still not sure how the double parenthesis fix got clobbered :thinking: 

In any case, those last two nits should be fixed

docs/admin/installation/hardware.rst
```diff
@@ -75,7 +75,12 @@ and recycled ThinkPad laptops work well for the
 *Admin Workstation*/*Journalist Workstation*.
 
 Please note that very old laptops or desktops may not work for the workstations.
-Since the release of Tails 3.0, 32-bit computers are no longer supported.
+Since the release of Tails 3.0, 32-bit computers are no longer supported.asdfasdf
```
eloquence, line 78: > asdfasdf

I've been there. ;-)

nathandyer: Not sure whether to laugh or cry :upside_down_face:

Starting to think this PR is cursed!

PR #734, opened 2025-11-20
Build a 2.12 version of docs

legoktm:
Since the 2.13.0 docs will be very different, keep a version of the old docs around for reference in case people need to fix things before they upgrade.

This will use the latest 2.12.x tag in case future updates/fixes need to be made to that branch.

## Test plan

* [x] Run `podman build --build-arg GIT_BRANCH=$(git rev-parse HEAD) --file ./deploy/Dockerfile . -t sd-docs`
* [x] Then `podman run -p 5080:5080 --rm sd-docs`
* [x] Now visit http://localhost:5080/en/2.12/ which should be a 2.12.x build. You can verify it has older content by checking http://localhost:5080/en/2.12/getting_support.html which should still reference the Support Portal and not Signal yet.

## Checklist

This change accounts for:


nathandyer:
Last week I looked into this further and could not find a cause for what I was seeing.

I chatted with @legoktm about this just a few moments ago. The oddities with the links in the selector appear to only be happening on my system. He verified he can build it and it works just fine.

We feel OK with going ahead and merging, and _if_ the issue persists we can troubleshoot further in a follow-up.

PR #735, opened 2025-11-21
Changes for 2.13.0 (securedrop-admin as Debian package)

### LINKED ISSUE #733: Update documentation installation/upgrade of version 2.13.0 (Debian package)
In SecureDrop 2.13.0 `securedrop-admin` is a Debian package. See https://github.com/freedomofpress/securedrop/pull/7606

Consequently the instructions for installing SecureDrop need to change. 

Upgrade instructions to 2.13.0, which workarounds for potential issues, also need to be included.

ChumOfChance:
Fixes #733 

With the move to `securedrop-admin` as a Debian package, there are numerous changes throughout the docs. The paths and commands for most operations have changed. 

Please closely check all the paths that have changed. These changes are based on https://github.com/freedomofpress/securedrop/pull/7606#issuecomment-3542424304 and https://github.com/freedomofpress/securedrop-docs/issues/733#issuecomment-3558722529 but it's possible I made a mistake or assumed the wrong path. I think it would be wise to test the documentation steps for many/all of these scenarios. 

Note that we will keep a 2.12.10 version of the docs live at least for some time after the 2.13.0 release, as a reference for users using previous versions who are troubleshooting issues before upgrading or need to migrate/restore from backups on older versions.

I've pre-emptively linked to where I expect the blog post announcing 2.13.0 to be posted, assuming it will have more details about the move to a Debian package. 

There are also some minor edits for consistent style or errors as I noticed them.

Major changes:

- The cloned `securedrop` repository is only needed for the initial bootstrapping process. I have changed the Install instructions so that this cloning is done to `/~` rather that to `~/Persistent`. The repo will not persist after reboot on Tails. I believe it is no longer needed once the `securedrop-admin` package is installed successfully. This is intended to ease confusion for long-time users who may revert to the habit of trying to run `./securedrop-admin` from the `~/Persistent/securedrop` directory and reinforced the change. 

- I removed the https://docs.securedrop.org/en/stable/admin/maintenance/update_workstations.html page because it is no longer relevant. 

- I have re-ordered one part of the Install instructions, moving the initial `securedrop-admin trailsbootstrap` run on the Admin Workstation, to an earlier step than it would be if following the older instructions closely. This is mostly done to simplify the instructions for setting up a Journalist Workstation. 

	In this PR, the instructions to "Onboard Journalists" (create a Journalist Workstation) reference sections of the "Set up the Admin Workstation" page, which is not ideal (this was the case previously as well). 

	Eventually I think it would be better to split "Set Up the Admin Workstation" into a page with Admin Workstation-specific instructions, and one with generic instructions for setting up a Tails USB with the `secredrop-admin` package, since this is done for both Admin Workstations and Journalist Workstations.

	Alternatively, the instructions could be repeated in the "Onboard Journalists" page. 

	I'm opting not to make such changes in this PR, instead sticking closer to the status quo, because I think making such change properly coherent would necessitate bigger changes to the entire flow of instructions for installing and deploying. 

## Test plan
The following pages/sections have instructions or troubleshooting steps that now have different commands and/or paths. These should be checked and in many cases tested.
- [ ] Upgrade instructions 2.12.10 to 2.13.0 (+ Manual Upgrade)
- [ ] Full Fresh Install (many pages)
- [ ] Offboarding
- [ ] Onboarding Admins
- [ ] Onboarding Journalists
- [ ] SSH Over Local Network
- [ ] Tor POW
- [ ] Backup and Restore
- [ ] Backup Workstations
- [ ] Kernel Troubleshooting
- [ ] Rebuild Admin Workstation
- [ ] Updates Over Tor
- [ ] FAQ
- [ ] SecureDrop Admin Utility
- [ ] HTTPS on Source Interface
- [ ] Test the Installation
- [ ] OSSEC Alerts

### DIFF
--- a/docs/upgrade/2.12.10_to_2.13.0.rst
+++ b/docs/upgrade/2.12.10_to_2.13.0.rst
@@ -5,9 +5,14 @@ Upgrade from 2.12.10 to 2.13.0
 Changes in SecureDrop 2.13.0
 ----------------------------
 
-SecureDrop 2.13.10 marks a major change from previous versions as the ``securedrop-admin`` tool is now provided as a Debian package installed on Tails using ``apt`` and available system-wide, rather than as a tool that must be run directly from a local cloned version of the SecureDrop repository. 
+SecureDrop 2.13.10 marks a major change from previous versions: the ``securedrop-admin`` tool is now provided as a Debian package installed on Tails using ``apt`` and available system-wide, instead of being run directly from a cloned version of the SecureDrop Git repository. 
 
-After you have successfully performed this upgrade on an *Admin Workstation*, the ``~/Persistent/securedrop`` directory is no longer used. Configuration files for your SecureDrop are moved to ``~/.config/securedrop-admin`` and stored there going forward. Running ``securedrop-admin`` commands is different two important ways: you no longer need to run them from the ``~/Persistent/securedrop`` directory and you should not include a leading ``./``. 
+Some key changes you should be aware of:
+
+- The ``~/Persistent/securedrop`` directory is no longer used. 
+- The ``securedrop-admin`` utility is installed in the default path and can be run as ``securedrop-admin <command>`` in a terminal window.
+- Configuration and backup files will now be stored in ``~/.config/securedrop-admin``.
+- The GUI updater is no longer required.
 
 For more details, see the `announcement for this release <https://securedrop.org/news/securedrop-2_13_0-released/>`_. 
 


docs/upgrade/2.12.10_to_2.13.0.rst
```diff
@@ -0,0 +1,80 @@
+
+Upgrade from 2.12.10 to 2.13.0
+==============================
+
+Changes in SecureDrop 2.13.0
+----------------------------
+
+SecureDrop 2.13.10 marks a major change from previous versions as the ``securedrop-admin`` tool is now provided as a Debian package installed on Tails using ``apt`` and available system-wide, rather than as a tool that must be run directly from a local cloned version of the SecureDrop repository.
```
nathandyer: I _think_ local is an adverb here, so it should be:

```suggestion
SecureDrop 2.13.10 marks a major change from previous versions as the ``securedrop-admin`` tool is now provided as a Debian package installed on Tails using ``apt`` and available system-wide, rather than as a tool that must be run directly from a locally cloned version of the SecureDrop repository. 
```

Then again, I don't pretend to have an expert grasp on the English language, so feel free to dismiss if I'm horribly mistaken here.

docs/admin/maintenance/backup_and_restore.rst
```diff
@@ -59,17 +59,10 @@ contact us through the `SecureDrop Support Portal`_.
...
-
-Run ``git describe --exact-match`` to verify that your workstation is running
-the latest version of  SecureDrop, |version|. If not, you should update it
-before proceeding.
-
 Check Connectivity
 ''''''''''''''''''
 
-Verify that your *Admin Workstation* is able to run Ansible and connect to
+Open a **Terminal** on your *Admin Workstation* and verify it is able to run Ansible and connect to
```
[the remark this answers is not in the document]
ChumOfChance: I agree. It shouldn't be bold. There isn't an agreed convention, but the latest commit should give the full menu invocation of the Terminal the first time it is used in each *section* on a page.

docs/admin/reference/securedrop_admin.rst
```diff
@@ -45,22 +44,23 @@ There are two primary reasons why you may want to update the system configuratio
...
-   SecureDrop <https://securedrop.org/news/release-announcement/>`__. Apply all
-   available updates, including for the Tails operating system.
-#. Run ``./securedrop-admin sdconfig``. This will display the current
+#. Determine the current version of `securedrop-admin` you have installed on your *Admin Workstation* by running:
+
+  .. code:: sh
+
+    apt-cache policy securedrop-admin
+
+  If the version is the same as the version number displayed in the footer of your *Source Interface*, you are running the applicable version of the SecureDrop code on your workstation, and can proceed to the next step.
```
jskinne3: Would this benefit from a screenshot?

ChumOfChance:
I've made additional changes based on a few things noticed during hands on testing of rc-1 (https://github.com/freedomofpress/securedrop/issues/7721)

I have not stepped through the backup/restore or offboarding pages yet

docs/admin/reference/securedrop_admin.rst
ChumOfChance: The output of this command is fairly clear. The problem with a screenshot or quoting terminal output here is that it will become out of date with future versions. I prefer a more "evergreen" approach than giving examples which will show an outdated version as soon as a new version is released.

PR #747, opened 2026-01-09
change off-boarding checklist to Signal

### LINKED ISSUE #738: Add "signal" to offboarding steps
## Describe the change
The [off-boarding checklist](https://docs.securedrop.org/en/stable/admin/deployment/offboarding.html#off-boarding-checklist) mentions the support portal, but given the [switch to supporting users over signal](https://securedrop.org/news/changing-how-we-deliver-support/) this should be reflected in the offboarding section. It's possible it's also missing in the on-boarding section or similar, but I haven't checked.

Specifically [this section](https://github.com/freedomofpress/securedrop-docs/blob/5dd6f36/docs/admin/deployment/offboarding.rst?plain=1#L24-L26):

> Inform the [SecureDrop Support](https://github.com/freedomofpress/securedrop-docs/blob/5dd6f365fcb3310b0cd65d662d63b5c9f82327ea/docs/admin/deployment/offboarding.rst#id7)) team that the user's support portal account should be deactivated, and indicate if any new staff members should be added.

It's pretty minor as we know what to do once reached out, but for completeness sake we could adjust the language there to say "support" accounts, in general, rather than "support portal".

## Additional context

1.

ChumOfChance:
Fixes #738 

Small change to the off-boarding checklist, to reflect that we now use Signal for support instead of the previous support portal.

### DIFF
--- a/docs/admin/deployment/offboarding.rst
+++ b/docs/admin/deployment/offboarding.rst
@@ -21,8 +21,7 @@ important to off-board them from SecureDrop.
 Off-boarding checklist
 ----------------------
 
-- :ref:`Inform the SecureDrop Support <getting_support>` team that the user's
-  support portal account should be deactivated, and indicate if any new staff
+- :ref:`Inform the SecureDrop Support <getting_support>` team that the user should be removed from any support Signal groups, and indicate if any new staff
   members should be added.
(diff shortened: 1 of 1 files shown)


nathandyer (approved):
LGTM!

Issue #749, opened 2026-01-14
Documentation Cleanup / Restructuring


nathandyer:
This is a meta ticket to track ongoing work in the near future related to a wider SecureDrop docs restructuring / cleanup effort.

In particular, we're planning to focus on the following items:

- [ ] Integrating SecureDrop Workstation documentation within the main project documentation
- [ ] Making the "index" page of the docs more inviting to people unfamiliar with the project
- [ ] Removing the need to jump around as much within individual documentation pages ("choose your own adventure")
- [ ] Generally, making the docs more concise and approachable

This will be a multi-sprint effort, likely involving myself and @ChumOfChance, along with reviews from @jskinne3 and the rest of the SecureDrop team.

deeplow:
I second the "workstation" nomeclature clarification.

> I do advocate for having a a page with some Qubes introductory material would be really useful.

I agree with the usefulness for admins, but wanted to understand a bit more how much we expect this to overlap with regular users. I shared some thoughts in https://github.com/freedomofpress/securedrop-workstation-docs/issues/344#issuecomment-3773065212.

deeplow:
Minor nit / follow up from another issue. **TD;DR** when doing the renames, also do it for `starting_client.rst` -> `starting_inbox.rst`.

> > Internal notes is a fair point. Combining the workstation and regular docs as part of [freedomofpress/securedrop-docs#749](https://github.com/freedomofpress/securedrop-docs/issues/749) is already going to change a lot of file names/locations where certain information is found in the docs (see the `main` branch of https://github.com/freedomofpress/securedrop-docs). So I think either now, or during that restructure, changing the name of `starting_client.rst` should happen.
>
> I've been thinking about this thread for a bit, and I believe waiting for the restructure makes the most sense.
> 
> Once we complete the docs restructure, there won't _be_ a Sphinx documentation site at workstation.securedrop.org, and _everything_ will need to redirect to the main docs site. So implementing the redirects at that level (with infras help) rather than using the options within Sphinx is probably the more future-proof way to go.
> 
> And since this will be happening in the relatively short term (once admin tools on Qubes lands), I think it probably makes sense to _not_ rename that file yet, but handle all the wonkiness from the rename as part of that larger transition.
> 
> My preference would be to go ahead and merge this as-is, then, if the reviewers are OK with that.
> 
> — _Originally posted by @nathandyer in https://github.com/freedomofpress/securedrop-workstation-docs/issues/398#issuecomment-4729437429_

PR #770, opened 2026-02-25
Create a foundation for post-App, post-Tails docs

nathandyer:


nathandyer:
This should be ready for review later today, I just need to fix a couple remaining issues with references to the Transfer Device (which will no longer need to exist in a post-Tails world)

nathandyer:
This isn't perfect, but if I keep poking at it we'll never have a firm foundation to build from.

ChumOfChance:
Starting to review this. My approach so far is to make sure we are not accidentally deleting anything we might/will need. There are some entire pages which are being deleted which we will need to re-create. 

I am thinking of documenting those and adding them as issues to https://github.com/freedomofpress/securedrop-docs/issues/749 (e.g. create new "offboarding" pages)

nathandyer:
> I am thinking of documenting those and adding them as issues to #749 (e.g. create new "offboarding" pages)

Thanks, I think this sounds like a good plan! If I have removed a large swath of content that should either be put back or more carefully picked from, let me know and I'll be happy to do so up-front in this PR; that said, if there are ones that would be better/faster to fully recreate from scratch given how much needs to be changed, those are likely good candidates for those additional tickets.

docs/journalist/workstation.rst
```diff
@@ -1,189 +0,0 @@
...
-admin for assistance.
-
-|Journalist account profile|
-
-.. include:: ../includes/tor-security-setting.txt
-
-.. _daily_journalist_alerts:
-
-
-Daily Journalist Alerts About Submissions
```
ChumOfChance: This is the only place the Daily Alerts are really explained, which should probably be preserved. There is a link to the Journalist section "index" page from https://docs.securedrop.org/en/stable/admin/installation/install.html#configure-the-installation where setting up this feature is done. The explanation could be moved there, or to the new SDW-specific journalist/journalist page or somewhere in that section?

docs/admin/maintenance/rebuild_admin.rst
```diff
@@ -1,481 +0,0 @@
-Rebuilding an *Admin Workstation* USB
```
ChumOfChance: We should keep this page, and adapt it for working from the Admin VM on Qubes/SDW.

docs/journalist/workstation.rst
nathandyer: Great flag. I thought about adding it at the configuration page you linked, but I think it makes sense to keep this part more journalist-facing. I've included it in the FAQ for now, and will be interested to hear if you think that is a reasonable spot.

docs/admin/deployment/remote.rst
```diff
@@ -1,224 +0,0 @@
...
-of which have desktop apps, instead of more common tools like Slack or
-unencrypted email.
-
-For security reasons, we advise against taking photos of documents using your
-phone, but if you decide to do so, please `see our guide to taking private
-photos with Signal 
-<https://freedom.press/training/taking-private-photos-signal/>`_.
-
-
-Protecting, Moving, or Taking Down Your SecureDrop Instance
```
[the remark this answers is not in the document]
nathandyer: Added it to the decommissioning guide, and turned the guide into two sections (basically, "temporary" and "permanent"). Let me know what you think of that.

docs/admin/maintenance/rebuild_admin.rst
nathandyer: I've added the page back in as a placeholder; will leave the re-write as an extra follow-up task.

docs/admin/deployment/remote.rst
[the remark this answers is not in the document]
ChumOfChance: Nice! Yes I think that's a good spot.

ChumOfChance (approved):

PR #800, opened 2026-04-15
Bump version to 2.15.0 and add upgrade guide

nathandyer:
## Test plan
- [x] Visual review
- [x] CI passes

## Checklist

This change accounts for:
- [x] local preview of changes beyond typo-level edits

PR #803, opened 2026-04-23
docs: fix typo in upgrade guide from v2.14.0

cfm:
Fixes a tiny typo in the new version numbers introduced in #800.


nathandyer (approved):
Can't believe we all missed that one!

PR #806, opened 2026-05-21
Proposed dynamic copyright date

jskinne3:
## Test plan
Ensure `2015-2026` appears at the bottom of any page


nathandyer (approved):
This is a great idea! Never noticed the Copyright years were so out of date.

I tested this and it looks/works as expected.

PR #808, opened 2026-06-11
Standardize and polish for docs restructuring

### LINKED ISSUE #790: Standardize formatting and conventions
The docs need some polish:
- Implement consistent capitalization for page titles and headings (Title Case or not?)
- Standardize various terms (e.g. e-mail vs email)

Internally, we could also standardize the naming pattern for rst label anchors e.g. `daily_journalist_alerts` vs `Daily Journalist Alerts`.

ChumOfChance:
Fixes #790 

The goal of this PR is to apply some style and formatting standardization, as well as polish, to the current `main` branch. Our docs already had inconsistent application of styles and terms and this has been exacerbated by combining two previously distinct documentations. This branch is slated to be published as a major restructuring of the docs when we hit a version of SecureDrop that is fully Qubes-based. 

- [x] Standardize use of terms
  - "email" over "e-mail" following AP style and similar small changes
  - I've applied the preferred term "USB flash drive" as opposed to the myriad of different options that were being used, based on Wikipedia's preferred term: https://en.wikipedia.org/wiki/USB_flash_drive
  - "*Onion Service*" as per our glossary, "onion address" as per the Tor documentation, "onion name" or "short onion name".
- [x] Remove use of Title Case in headers and subheaders in accordance with FPF's preferred style
- [x] Update Glossary and consistently apply italics to ~*Glossary Terms*~ terms of art in the text
  - A few terms like "Admin Interface" were previous title case and italicized like other glossary terms, despite not appearing in the glossary. I've applied that formatting consistently to all such "terms of art" that were previously italicized + title case, regardless of whether they currently or ever did appear in the glossary. 
  - There are few exceptions, for example "journalist accounts" (accounts of the journalist type) and introductory text referring to journalists and sources in more general terms
  - Restored or partially restored a few terms to the glossary I had previously removed
  - I've tried to steer this PR away from what should be in the glossary/formatted as a "term of art". I think there are few discussions needed there and probably another PR down the line. In this case, I am just trying to apply consistent formatting and style.
- [x] Replace remaining references to "App" with "Inbox"
  - Also cleaned up other references to "Application" which could be confusing
- [x] Fix whitespace
  - Made the use of extra lines around headers consistent, which improves readability of the source
- [x] Fix frequent use of U+2019 (right quotation mark) in place of apostrophe

I intentionally skipped over the training schedule, as it is currently on the way to removal https://github.com/freedomofpress/securedrop-docs/pull/809


docs/admin/deployment/onion_name.rst
```diff
@@ -1,14 +1,14 @@
...
+What are onion names?
 ^^^^^^^^^^^^^^^^^^^^^
 
 Onion names are short, memorable addresses that visitors can use to access an 
-onion service (e.g., a news organization's SecureDrop) using Tor Browser. 
+*Onion Service* (e.g., a news organization's SecureDrop) using Tor Browser. 
 
 Imagine a SecureDrop instance for a new organization called
-*The New York World* with a .onion address like this:
+*The New York World* with an onion address like this:
```
nathandyer, line 11: I wonder if omitting the `.` here makes it less clear that the address ends in `.onion`?

docs/admin/deployment/onion_name.rst
```diff
@@ -20,11 +20,11 @@ The general format for a SecureDrop onion name is:
 
 ``<organization>.securedrop.tor.onion``
 
-How They Work
+How they work
 ^^^^^^^^^^^^^
 
 Onion names are supported in the desktop version of Tor Browser (introduced 
-in version 9.5). The mapping between onion names and the full-length .onion 
+in version 9.5). The mapping between onion names and the full-length onion
```
nathandyer, line 27: Same question/concern as before re: omitting the `.`

docs/admin/installation/create_admin_account.rst
```diff
@@ -1,7 +1,7 @@
-Create an Admin Account on the *Journalist Interface*
+Create an admin account on the *Journalist Interface*
```
nathandyer, line 1: For consistency, does `admin` need to be capitalized and italicized as a glossary term?

docs/admin/installation/create_admin_account.rst
```diff
@@ -1,7 +1,7 @@
...
 =====================================================
 
-In order for any user (admin or journalist) to access the
+In order for any user (admin or *Journalist*) to access the
```
nathandyer, line 4: Same concern re: admin being a glossary term

docs/admin/installation/create_admin_account.rst
```diff
@@ -10,29 +10,27 @@ In order for any user (admin or journalist) to access the
...
-running a command on the *Application Server*. After that, the Journalist
-Interface admin can create additional accounts through the web
-interface.
+running a command on the *Application Server*. After that, the admin can create additional accounts through the web-based *Journalist Interface*.
 
 .. _Create Admin CLI:
 
 To create an admin account via the command line,
-:doc:`SSH to the Application Server <../installation/test_the_installation>`,
+:doc:`SSH to the*Application Server* <../installation/test_the_installation>`,
```
nathandyer: We lost a space along the way

```suggestion
:doc:`SSH to the *Application Server* <../installation/test_the_installation>`,
```

docs/admin/installation/create_admin_account.rst
```diff
@@ -55,7 +53,7 @@ Passphrases include the spaces between the words, but not leading or trailing
 whitespace. Be sure to save this passphrase in the appropriate KeePassXC database.
 
 Once that's done, you should open Tor Browser |TorBrowser| and
-navigate to the *Journalist Interface*'s .onion address. Verify that you
+navigate to the *Journalist Interface*'s onion address. Verify that you
```
nathandyer, line 56: Just flagging this as a spot where we'll need to update if we decide to include the `.`

docs/admin/installation/install.rst
```diff
@@ -286,20 +286,20 @@ an email to securedrop@freedom.press.
...
+- ``app-sourcev3-ths`` contains the v3 onion address of the *Source
   Interface*.
-- ``app-journalist.auth_private`` contains the ``onion`` address and private key
+- ``app-journalist.auth_private`` contains the onion address and private key
   providing access to the *Journalist Interface*.
-- ``app-ssh.auth_private`` contains the ``onion`` address and private key
+- ``app-ssh.auth_private`` contains the onion address and private key
   providing SSH access to the *Application Server*.
-- ``mon-ssh.auth_private`` contains the ``onion`` address and private key
+- ``mon-ssh.auth_private`` contains the onion address and private key
```
nathandyer, line 302: Flagging for if we want to change the `.`

docs/admin/installation/provisioning_usb.rst
```diff
@@ -1,24 +1,24 @@
-Provisioning Export USB devices
-===============================
+Provisioning USB *Export Devices*
```
nathandyer, line 1: I wonder if saying USB is redundant here? 

```suggestion
Provisioning *Export Devices*
```

docs/admin/reference/ssh_access.rst
```diff
@@ -1,8 +1,9 @@
 Logging in via SSH
 ==================
 
-SSH Over Tor
+SSH over tor
```
nathandyer: Tor should continue to be capitalized

```suggestion
SSH over Tor
```

docs/admin/reference/ssh_access.rst
```diff
@@ -113,21 +114,21 @@ into each server (via ``ssh app`` and ``ssh mon``) and run the following command
...
+--------------------
 
-Adding Users (CLI)
+Adding users (CLI)
 ^^^^^^^^^^^^^^^^^^
 
 After the provisioning of the first admin account, we recommend
-using the Admin Interface web application for adding additional journalists
-and admins.
+using the Admin Interface web application for adding additional journalist
```
nathandyer:
```suggestion
using the *Admin Interface* web application for adding additional journalist
```

docs/includes/backup-and-update-reminders.txt
```diff
@@ -1,15 +1,15 @@
-Back Up the Tails Workstations
+Back up the tails workstations
```
nathandyer: Tails should stay capitalized

```suggestion
Back up the Tails workstations
```

docs/introduction/securedrop_workstation.rst
```diff
@@ -159,29 +161,30 @@ For more about the security features of Qubes, see
...
 application) can access these other files, and may be able to exfiltrate them.
 
 In spite of the air-gap, this may be possible through physical channels used to transfer files
-off the SVS (e.g., USB drives), or by motivating the journalist user to perform an
+off the Secure Viewing Station (e.g., USB flash drives), or by motivating the *Journalist* to perform an
 unsafe action (e.g., `scanning a QR code <https://securedrop.org/news/security-advisory-do-not-scan-qr-codes-submitted-through-securedrop-connected-devices/>`__).
 
-Because the air-gapped SVS has no Internet access, updates can only be performed using
-another computer and a USB drive. In practice, newsrooms may not update their SVS
+Because the air-gapped Secure Viewing Station has no Internet access, updates can only be performed using
```
nathandyer:
```suggestion
Because the air-gapped *Secure Viewing Station* has no Internet access, updates can only be performed using
```

nathandyer: This is one instance, but this whole section needs to be updated since that's a glossary term

docs/appendices/glossary.rst
```diff
@@ -1,22 +1,36 @@
...
+The *Application Server* runs the SecureDrop server application. This server hosts both
+the website that *Sources* access (the *Source Interface*) and the website that
+*Journalists* access (the *Journalist Interface*). Both are published
+through an *Onion Service* because *Sources*, *Journalists*, and admins
 may only connect to this server using Tor.
 
+Export Device
+-------------
+
+The *Export Device* is the physical media (e.g., designated USB flash drive) used to transfer decrypted documents from the Secure Viewing Station to a journalist's everyday workstation, or to another computer for additional processing.
```
nathandyer:
```suggestion
The *Export Device* is the physical media (e.g., designated USB flash drive) used to transfer decrypted documents from the *Secure Viewing Station* to a journalist's everyday workstation, or to another computer for additional processing.
```

docs/appendices/threat_model/threat_model.rst
```diff
@@ -671,7 +671,7 @@ What a Global Adversary Can Achieve Against the Source, Admin, or Journalist:
    to spoof an organization's HTTPS *Landing Page*, thereby tricking the
    source into visiting a fake SecureDrop site.
 
-What a Random Person on the Internet Can Achieve
+What a random person on the internet can achieve
```
nathandyer, line 674: We should double-check, I _think_ Internet still needs to be capitalized

docs/appendices/threat_model/threat_model.rst
```diff
@@ -453,7 +453,7 @@ What a Physical Seizure of the Admin's Property Can Achieve
    volume, password database, and two-factor authentication device will
    allow the attacker to access both servers and the *Journalist Interface*.
 
-What a Compromise of the Journalist's Property Can Achieve
+What a compromise of the journalist's property can achieve
```
nathandyer: An instance of inconsistent capitalization of "Journalist" that we'll want to be careful about

nathandyer (changes requested):
These changes are looking great! I know this was a massive undertaking, thank you so much @ChumOfChance

A handful of things I noticed as part of this first review pass:
* SecureDrop Inbox should be added as an official glossary term, and treated as such throughout. 
* Journalist is not consistently italicized as a glossary term. Sometimes it is, sometimes it isn't.
* I thought I recalled us deciding that Administrator would be part of the glossary to match the other user roles. (I may be mistaken). Either way, I think it _should_ be, and treated the same as Journalist or Source when necessary.
* Sometimes the random capitalization in the middle of a sentence when a glossary term pops up feels a little jarring. I'm wondering if the italicization is sufficient, and requiring them to be capitalized every time is overkill?
* We should add the glossary terms from the TODOs as part of this PR
* I see we're looking to standardize on the term "USB Flash Drive". I'm curious if that's the most user-friendly term. At first glance, "drive" probably doesn't mean much to most people (I vaguagely recall we had this discussion in years past and settled on "device"). Flash is likewise only meaningful to some. I tried to find a previous conversation we had about this, but was unable to do so. I know at once point in time we had terms like "thumb drive" mixed in, and I think in general we started just using "USB drive," although that's less descriptive because it doesn't clarify between a flash drive vs a portable hard drive/SSD, etc., and it is also problematic because it uses the word "drive" like I mentioned above. Maybe there isn't a perfect term? In any case, I'm fine with using "USB Flash Drive," I just want to make sure we're intentional about it.
* Do we feel that using "onion addresses" across the board, and not specifying the `.` part to emphasize that addresses end in `.onion`, will be clear enough to end users? (Probably so, just asking)

ChumOfChance:
I forgot to mention an important thing which is that I intentionally tried to not make change to the threat model page beyond the title case changes. The threat model feels like an external document for reference, it's authored outside of the rest of the docs (and I think expects a big revision soon to bring it up to date with a fully Qubes-based SecureDrop). Being published inside the docs I felt removing title case from the headers was needed in this PR, but I didn't want to go through and fuss with all the glossary terms (there are also a lot).

But I can see how that will make this PR cleaner, even if it might be changed a bit later!

ChumOfChance:
Re: the glossary, I planned to do a subsequent PR that specifically addresses the glossary fixing https://github.com/freedomofpress/securedrop-docs/issues/814 and https://github.com/freedomofpress/securedrop-docs/issues/813 ; few of the definitions need to be updated, some added, and then apply the italicization to the new glossary terms.

I wanted to keep that work out of this PR, and focus this more tightly on style/standard terms/formatting (it's already so massive). I only added back two glossary terms that were already mostly italicized in the text, and adding them back should be uncontroversial (Admin and Journalist Workstation are terms we are reusing with new definitions, and they were removed from the glossary in a previous PR by me I think, prematurely). This PR should be consistent (everything in the glossary is italicized).

ChumOfChance:
Re: ".onion", I think ".onion address" is not a good term to use. Onion services are reached using an onion address, which is a long string of characters ending in ".onion". I think it's better to describe that relationship coherently when needed (like in the glossary), and I think users will understand. I think we should definitely not use ".onion address" every time it appears in the documentation, and previously we had a mishmash of many different terms. 

When it is necessary, we should point out explicitly that an onion address ends in ".onion", or that it's important to include the ".onion" when copying and pasting an onion address. There are a few examples in the docs, e.g. in the source documentation I left in:

> Using Tor Browser, find the ".onion" address for the SecureDrop for
the organization that you wish to submit to.

Which I think is a good way to succinctly communicate to the a source user in a very non-technical way (but I think the quotes are important here!).

I did notice an errant `.onion` which I think should be changed to regular quotes while searching for the above example though!

ChumOfChance:
I agree with you about the capitalization + italicization. I think dropping the title case from glossary terms when they appear would be a good chance, if you think we can make that call I'm game to add it to this PR.

ChumOfChance:
Saving the most controversial for last :laughing: re: "USB flash drive". We had so many different terms for this and I am sure it has caused confusion for users. I went to Wikipedia to settle the issue, which prefers "USB flash drive" ("also known as a thumb drive"). If there is a better precedent or resolved discussion for SecureDrop's preference I can abide that rather than opening the can of worms, but of the [available options](https://en.wikipedia.org/wiki/USB_flash_drive#cite_note-3) I'm inclined to agree with Wikipedia. 


It is a mouthful, but it's descriptive. If we think it's too wordy, I think a compromise would be to invoke "USB flash drive" when being explicit, the first time it is mentioned in a page/section, and then more often shortening to "drive" thereafter in the same page/section. This will give us things like Admin Workstation drive, Tails drive, Qubes installation drive appearing after "USB flash drive" has already made it explicit that this is referring to one of those little USB stick keychain thumb-sized things.

docs/appendices/threat_model/threat_model.rst
ChumOfChance: I think it should be, personally, but that is sadly an outlier position of people like us. AP stylebook, which is what FPF follows, says it should not be. See https://www.ap.org/the-definitive-source/products-and-services/ready-to-lowercase-internet-and-web/

docs/introduction/securedrop_workstation.rst
ChumOfChance: Secure viewing station is actually not in the glossary anymore. The SVS will be a legacy item only relevant to people migrating, so I don't *think* it should appear in the glossary? Instead the title case here should be removed. 

I also changed many cases of "SVS" to "secure viewing station" since I think we should be avoiding acronyms like that in the documentation (especially when we will be referencing something from the previous gen of SecureDrop. Looks like I missed a few though!

docs/includes/backup-and-update-reminders.txt
ChumOfChance: Yes!

docs/admin/reference/ssh_access.rst
ChumOfChance: Yes! Thank you.

docs/admin/installation/provisioning_usb.rst
ChumOfChance: Hmm.. I think it's good to keep USB here. "Export Device" is kind of a vague term out of context. There are two ways currently to export materials a USB flash drive (the Export Device) and printer. I was thinking at the header of this section it would be good to keep "USB", (as it appeared in the previous header). A printer is also attached via USB, but I still think it's a good cue for the reader. I'm not strongly attached to it though.

docs/admin/reference/ssh_access.rst
ChumOfChance: Should "Admin Interface" be italicized her? It isn't in the glossary. 

I do think it would be worth considering adding the Journalist/Admin/Source web interfaces to the glossary to help clarify all the different... things, but presently I'm not sure how that will shake out in the post-Tails SecureDrop. My understanding is likely the Admin Interface will definitely still exist as a web interface, but journalists may or may not still be able to use it? But I don't think we are sure yet and it might be worth holding off.

ChumOfChance: I'm realizing how many italicized instances there are of Admin Interface and Journalist Interface. I am going to clean that up as best I can.

ChumOfChance: "Admin Interface" has never been in the glossary from what I can tell! Which doesn't 100% strictly mean it shouldn't be in italics, but probably more means it should be in the glossary... I've opened https://github.com/freedomofpress/securedrop-docs/issues/818 because I think either way it's clear there is some work beyond the scope of this PR for my above comment.

ChumOfChance:
I did some more work and updated the PR description. @nathandyer I think in particular the glossary/italics stuff should be easier to review, sorry it was still a loose before. Other open questions remain but I think I've replied to all your review comments so far.

nathandyer:
Thanks @ChumOfChance, I'll do another pass and also reply individually to your inline comments/messages as time allows. Hoping to take another look today, but if not, I'll get to it early next week.

nathandyer (changes requested):
Thanks @ChumOfChance, these changes are looking great!

I definitely like saying Secure Viewing Station in place of SVS consistently, and treating it like a glossary term (although as you note the SVS is effectively on its way out with the transition to Qubes, so its days are numbered beyond a forthcoming Tails -> Qubes migration guide).

I noticed you changing "Daily Journalist Alert" to "daily journalist alert", presumably because it's not part of the glossary. I do think it's worth thinking about if it _should_ be, and potentially making it stick out more just so it's clear to users that it's its own special thing ™️ 

In any case, I feel like this is getting close to landing. Once you're back, let's find a time to chat through some of this (maybe synchronously?) and we can get this merged so we can build on top of it. With such a big change, I don't want to start making too many changes without this landing first.

nathandyer:
@ChumOfChance This all LGTM. Like I said yesterday, with this being such a large PR and with the nature of docs being what it is, I'm sure we'll find things here and there that we'll want to tweak. But this feels like it's in a good spot to merge.

There's one remaining merge conflict, which should be resolvable by just removing the training_schedule.rst file. If you'll take care of that (so I don't lock myself out from being able to approve and merge), I'll get this landed today.

nathandyer (approved):

PR #811, opened 2026-06-12
Point potential sources to howto.securedrop.tor.onion

### LINKED ISSUE #767: Consider whether or not we want to retain the Source Guide
We currently have a guide for Sources that walks through the process of using SecureDrop as safely as possible, but it has some potential op-sec footguns.

For the sake of argument, I'm wondering if it might make sense to refer sources to `howto.securedrop.tor.onion` instead?

nathandyer:
Fixes #767

## Test plan
- [ ] Visual review

## Checklist

This change accounts for:


docs/source/source.rst
```diff
@@ -1,22 +1,26 @@
 SecureDrop for Sources
 ======================
 
+.. warning:: This source guide is meant to illustrate the source submission process
+    for journalists and newsrooms, and is not intended to instruct sources how to
```
ChumOfChance: With this new warning here, the page is pretty top-heavy on warning/notes, filling the top half of the page on a laptop screen. Visually in a browser (especially a mobile browser) the effect is a bit overloading. I'm not sure if that's a plus in terms of someone actually reading this warning... but perhaps the following older, less-urgent, note could be moved to the bottom of the document?

docs/source/source.rst
```diff
@@ -1,22 +1,26 @@
...
+    If you are potentially interested in sharing information to a news organization
+    via SecureDrop, you should instead:
+
+    1. Start from a place with public Wi-Fi, like a coffee shop. Use a computer you own and control. Do not use a mobile phone or tablet. Never use a workplace computer or network.
+    2. Download, install, and open Tor Browser, then visit ``howto.securedrop.tor.onion``
+
+
 .. note::
 
-   This guide provides an introduction to using SecureDrop as a source.
```
legoktm: I would be okay removing this note and instead including it in the howto site somewhere.

docs/source/source.rst
legoktm: I do wonder if this note is better left on the docs homepage. How many sources will go directly here versus there first?

nathandyer: It's a good question! In an unscientific test (from a logged out Google account in a fresh browser profile), I searched for various things I can imagine a source might search (ex. "how do I submit something to a securedrop), and across the board the "How to submit" page came up first in the results.

I do think most sources would land here rather than the main docs.securedrop.org page, and because this information isn't as relevant for admins and sources, I'm afraid a warning like this might be more confusing than not on the main page.

That said, I agree completely that this page is really heavy on warnings/notes. I'll see if moving to the bottom helps, or if I can find another way to make it clear but not overwhelming.

docs/source/source.rst
nathandyer: I've filed a PR against the securedrop-info repo to include this note: https://github.com/freedomofpress/securedrop-info/pull/20/

nathandyer: Removed as suggested

docs/source/source.rst
nathandyer: I've force-pushed an updated commit that I believe addresses most of these concerns.

It offloads some of the warnings to howto.securedrop.tor.onion (see https://github.com/freedomofpress/securedrop-info/pull/20/)

It also moves part of the warning further down to "Before you begin...", and downgrades that particular portion from "warning" to "note" since it's informational.

ChumOfChance: With the now-note flagging that the source guide in the documentation is meant to illustrate the submission process for journalists and newsrooms, some confusing wording in the warning to potential sources jumps out at me. 

".. you should instead:" but instead of what is not clear. 

Maybe the note and warning should be combined in a single warning. 

"This source guide below is meant to illustrate the source submission process for journalists and newsrooms, and is not intended to instruct sources how to safely use SecureDrop. If you are potentially interested in sharing information to a news organization via SecureDrop, you should: 1,2"

docs/source/source.rst
```diff
@@ -41,6 +34,12 @@ Note that your Internet Service Provider, or ISP (e.g., Comcast/Xfinity,
 Cox, Wave, etc), may already have a record of your visit to this website,
 docs.securedrop.org.
 
+.. note:: 
+
+    This source guide below is meant to illustrate the source submission process
+    for journalists and newsrooms, and is not intended to instruct sources how to
+    safely use SecureDrop. 
+
 **Before you begin...**
```
ChumOfChance: This section on this page is a bit redundant with the same information provided in the warning at the top of the page. This information is also partially repeated on the before_you_submit page. I think my suggestion is to move/consolidate this to the "before_you_submit" page?

docs/source/source.rst
nathandyer: I like that suggestion!

docs/source/source.rst
nathandyer: Agreed, moved and consolidated as suggested

nathandyer:
This should be ready for another pass as well

docs/source/source.rst
```diff
@@ -1,34 +1,29 @@
...
+    1. Start from a place with public Wi-Fi, like a coffee shop. Use a computer you own and control. Do not use a mobile phone or tablet. Never use a workplace computer or network.
+    2. Download, install, and open Tor Browser, then visit ``howto.securedrop.tor.onion``
+    
 
 What is SecureDrop?
 -------------------
 
 SecureDrop is a tool that news organizations and NGOs use that enables secure
-and anonymous communication between whistleblowers and *Journalists*.
+and anonymous communication between whistleblowers and *journalists*.
```
ChumOfChance:
```suggestion
and anonymous communication between whistleblowers and *Journalists*.
```

docs/source/source.rst
```diff
@@ -1,34 +1,29 @@
...
 No personal information is collected; information submitted to SecureDrop is
 encrypted, and SecureDrop is not a “cloud” service. If you don't have sensitive
 information to send to a news organization, it may be okay to use a traditional
 methods such as phone or email when reaching out.
 
 SecureDrop can accept both messages and individual file uploads (up to 500MB).
-If you have multiple files to submit, you may do that. As a *Source*, you can also
+If you have multiple files to submit, you may do that. As a *source*, you can also
```
ChumOfChance:
```suggestion
If you have multiple files to submit, you may do that. As a *Source*, you can also
```

ChumOfChance (changes requested):

ChumOfChance (approved):

Issue #814, opened 2026-06-17
Completely remove references to V2 onion services.


ChumOfChance:
https://github.com/freedomofpress/securedrop-docs/pull/208 fixed https://github.com/freedomofpress/securedrop-docs/pull/208 by removing most references to V2 onion services. There are still a few remaining in the docs, notably in the Glossary or in reference to restoring from a backup of a system that was using V2 services. Given the [timeline](https://blog.torproject.org/v2-deprecation-timeline/) to depreciate V2 onion services, I think at this point we can completely remove any references to V2.

PR #821, opened 2026-06-26
Update Glossary

### LINKED ISSUE #813: Add Administrator to the glossary
## Describe the change

The documentation's glossary defines a list of terms, a few of which are technical (Onion Service, Two-Factor Authentication) and a few of which are concepts that are useful for defining roles of people in equipment in how SecureDrop works and is used. Historically this has included Journalist Workstation, Admin Workstation, Journalist, and Source.

We should add Administrator as a term to the glossary as part of the combining of the SecureDrop Workstation and "classic" documentation as we move away from the Tails-based system in favour of a fully Qubes-based SecureDrop. In this system just as an individual laptop may potentially be used as a Journalist Workstation, Admin Workstation, or both depending on how it is configured, so too might an individual be a Journalist, Administration, or both. Defining the Administrator role and privileges in the glossary alongside Journalist will help compare and contrast these roles, and help clarify which parts of the documentation are for which user-role, etc.

### LINKED ISSUE #818: Update references to journalist and admin interface for post-Tails
The documentation for SecureDrop Workstation references the Journalist Interface extensively. The authenticated onion service running on the server to which SecureDrop Inbox connects is technically called the "Journalist Interface" because this same onion service is what journalists have been connecting to in Tor browser on a Tails-based journalist workstation, where a web-GUI interface is presented. The "Admin Interface" used by administrators is this same web-based GUI, with a few additional administrative options available if logged in as an administrator.

At a organization using only Qubes-based SecureDrop Workstations, journalists will only use SecureDrop Inbox to interact with SecureDrop. The assumption is that there will no longer be a pre-configured way to access the web-based journalist interface in Tor browser on a workstation for journalists. If there is, it will be vestigial (perhaps undocumented) and not the primary way journalists should be interacting with SecureDrop. 

Administrators will still need to access this web-based interface to perform certain administrator actions like adding new users, at least for some time.

Looking to documentation for a post-Tails SecureDrop some of these terms need to change, but I'm not sure where to go exactly.

I don't think it makes sense to refer to the "Journalist Interface" when describing the authenticated onion service running on the server to which SecureDrop Inbox connects, and it's associated address and authentication credentials. This onion service/address which serves both the API and web-based GUI for journalists and administrators (distinct from the public source interface address) could perhaps be called the "Newsroom Address" or "Newsroom Service" or something similar. 

"Admin Interface" could refer strictly to the web-based GUI at this address, accessed in a browser. And "Journalist Interface" is there as a term if we are documenting how journalist can still access SecureDrop that way. 

Open to other suggestions or input on this!

ChumOfChance:
Fixes #813
Fixes #818

This PR updates the documentation Glossary, as well as how the glossary is integrated into the documentation. This should bring the glossary closer to a finished state for a post-Tails SecureDrop and establish some new standards for how to reference the glossary when writing/editing documentation.

- Various edits for clarity, accuracy, e.g. making how pub/priv keys are referenced consistent
- When glossary terms appear in the documentation, they are now styled only in Title Case (not *Title Case Plus Italics*). This is done because it's visually a bit nicer, and because rst provides no (easy) way to italicize link text (see next point)
- When a glossary term first appears in a page or section, it is linked to the relevant glossary entry when useful. I didn't do this for many appearances of Journalist or Source
- Add reference anchors for each entry for easy linking as per above
- Removes references to Onion v2 services (https://github.com/freedomofpress/securedrop-docs/issues/814) and adds explanation of onion address and authenticated onion services
- Updates/rewrites to several sections to reflect the Qubes/SecureDrop Workstation ecosystem
- Adds Administrator (#813)
- Adds SecureDrop Inbox and SecureDrop Workstation
- Removes Journalist Workstation from the glossary
  - A "SecureDrop Workstation" is the laptop used by journalists
  - An Admin Workstation is the laptop used by administrators
  - Less confusion with the current Tails-based system, less special terms/workstation overall
 - Adds Admin Interface, removes Journalist Interface (#818)
   - Admin Interface was sometimes previously erroneously styled as a glossary term despite not appearing in the glossary
   - Journalists will no longer use the web-based Journalist Interface to access SecureDrop, except perhaps as a fallback. That could be documented if we implement as it a backup access. However in the docs and glossary it makes more sense to refer to the Admin Interface, since administrators need to access this interface for user management and a few other features.
- Remove Two-Factor Authentication from the glossary. In 2026, everyone coming into contact with SecureDrop will most likely already be familiar with 2fa since it is so widely deployed in general
- Update the docs to reflect the new/changed/removed glossary terms

# Test Plan
- [ ] CI passes
- [ ] Visual review


docs/appendices/glossary.rst
```diff
@@ -1,151 +1,178 @@
...
 A number of terms used in this guide, and in the `SecureDrop workflow
 diagram <what_is_securedrop>`, are specific to SecureDrop. 
 The list below attempts to enumerate and define these terms.
 
+.. _glossary_administrator:
+
+Administrator
+-------------
+
+The Administrator installs, deploys, and maintains a SecureDrop instance. They connect to the Application and Monitor servers (using SSH) and to the Admin Interface (using Tor Browser) from an Admin Workstation.
```
nathandyer: I wonder if specifying "(using SSH)" is unnecessary here.

At least during initial setup, and then potentially later if something breaks, they will connect directly via keyboard and monitor.

docs/appendices/glossary.rst
```diff
@@ -1,151 +1,178 @@
...
 
 Application Server
 ------------------
 
-The *Application Server* runs the SecureDrop server application. This server hosts both
-the website that *Sources* access (the *Source Interface*) and the website that
-*Journalists* access (the *Journalist Interface*). Both are published
-through an *Onion Service* because *Sources*, *Journalists*, and admins
-may only connect to this server using Tor.
+The Application Server is a physical computer housed on-premises running the SecureDrop server application. Based on our :doc:`recommendations</admin/installation/hardware>`, this is typically a NUC or similar Mini PC. The Application hosts both
```
nathandyer:
```suggestion
The Application Server is a physical computer housed on-premises running the SecureDrop server application. Based on our :doc:`recommendations</admin/installation/hardware>`, this is typically a NUC or similar Mini PC. The Application hosts
```

Using "both" here doesn't really apply because the following list is more than two items.

docs/appendices/glossary.rst
```diff
@@ -1,151 +1,178 @@
...
 
-Journalist Alert Public Key
----------------------------
+.. _glossary_journalist_alert_key:
 
-The *Journalist Alert Public Key* is used for encrypting the daily alert
-that notifies *Journalists* via encrypted email about whether or not there has been
-submission activity in the past 24 hours. The *Journalist* uses an associated
-private key to decrypt the alerts.
+Journalist Alert Key
```
nathandyer, line 68: We refer to "Journalist Alert Key" here, but "Journalist Alert Public Key" further down. We should be consistent across the board.

docs/appendices/glossary.rst
```diff
@@ -1,151 +1,178 @@
...
 
 Landing Page
 ------------
-The *Landing Page* is the public-facing webpage for a SecureDrop instance. This
+
+The Landing Page is the public-facing webpage for a SecureDrop instance. This
 page is hosted as a standard (i.e. non-Tor) webpage on the news organization's
-site. It provides first instructions for potential *Sources* and includes
-the instance's :ref:`Source Interface <glossary_source_interface>` address.
+website. It provides first instructions for potential Sources and includes
```
nathandyer:
```suggestion
website. It provides instructions for potential Sources and includes
```

docs/appendices/glossary.rst
```diff
@@ -1,151 +1,178 @@
...
+
+.. _glossary_monitor_server:
 
 Monitor Server
 --------------
 
-The *Monitor Server* keeps track of the *Application Server* and sends out an
-email alert if something seems wrong. Only system admins connect
-to this server, and they may only do so using Tor.
+The Monitor Server is physical computer housed on-premises running the `OSSEC<https://www.ossec.net/>` intrusion detection software to monitor the Application Server and send out an email alert if something seems wrong. Like the Application server, it is typically a NUC or similar Mini PC. Only Administrators connect
```
nathandyer:
```suggestion
The Monitor Server is physical computer housed on-premises running the `OSSEC<https://www.ossec.net/>` intrusion detection software to monitor the Application Server and send out email alerts to notify Administrators about events, such as package installations or new login attempts. Like the Application server, it is typically a NUC or similar Mini PC. Only Administrators connect
```

Most of the alerts aren't sent when something is wrong, so we should better set expectations about what Admins will see

docs/appendices/glossary.rst
```diff
@@ -1,151 +1,178 @@
...
-Read more about `*Onion Services* in Tor's glossary
+Authenticated Onion Services require an additional authentication token be provided when connecting to the server. The Admin Interface, the API endpoint connected to by SecureDrop Inbox, and SSH access to the servers (in SSH-over-Tor mode) are all authenticated Onion Services.
+
+Read more about `Onion Services in Tor's glossary
 <https://support.torproject.org/onionservices/>`__.
 
-OSSEC Alert Public Key
+.. _glossary_ossec_alert_key:
+
+OSSEC Alert Key
```
nathandyer, line 120: Same feedback re: Journalist Alert [Public] Key. We should make sure these terms are consistent

nathandyer (changes requested):
I took some time for an initial pass here, and left some suggestions (mostly confined to the glossary page itself). These new additions were sorely needed!

The other changes all look OK to me (mostly just removing the italics from existing references). I really like the inclusion of linking to the glossary definition the first time a term is used.

I spotted a few instances where the references to the Journalist Interface were deliberately changed to Admin Interface. I think we need to be careful to make it clear that it's a fully separate area within the Journalist Interface. But I also know that Journalists won't be accessing that interface for much longer (given the move to Inbox), so there's an underlying question here of "for how long do we keep referring to that as the Journalist Interface?"

I do suggest we have at least one more reviewer on this PR (I know @jskinne3 was tagged here as well, his input would be valuable if his time allows). I also wonder if it's worth flagging the change in italics to the wider team.

docs/appendices/glossary.rst
ChumOfChance: I don't agree on this one. For glossary purposes, the raison d'etre for the Monitor Server is to host OSSEC to detect potential intrusions on the Application Server and alert administrators. 

We do not deploy OSSEC to alert admin about benign events. True, that is mostly what happens. But we have an admin reference separate page describing how to analyze and interpret OSSEC alerts and as long as OSSEC is being used, it's important for readers to understand the purpose of the Monitor Server. Setting expectations for admins for what to expect from OSSEC alerts in practice I think belongs elsewhere, maybe in introductory text for administrators, or post install maintenance (I think it's in there to some effect).

The phrase "something seems wrong" isn't great, but after naming it as "intrusion detection software" I thought it was OK to leave, but I think your suggestion does point to how this could be improved. See https://github.com/freedomofpress/securedrop-docs/pull/821/commits/5e5a25fa710992a716914a2d193ac6c73ccd2f48 I've made it more descriptive.

docs/appendices/glossary.rst
ChumOfChance: Previously the glossary had entries for "Journalist Alert Public Key" and "OSSEC Alert Public Key" which both referenced an associated but unnamed private key. The entry for "Submission Key" explicitly mentioned the Submission Public Key and Submission Private Key, both styled as sub-glossary terms. 

I've changed the former two entries to match the latter, where the main glossary entry is the _____ Key and then the ____ Public Key and Private Key are both mentioned in title case as terms that could be referred to elsewhere. I think it's important to do this for the alert keys because they are less sensitive but it is still a source of confusion that there is a public key and private key, that they are used for different things... without getting into a full PGP 101.  

Now that the email alerts page has better explanations of the creation of the keys and where they go (https://github.com/freedomofpress/securedrop-docs/pull/826), I added links to that page for both of the alert key glossary entries in https://github.com/freedomofpress/securedrop-docs/pull/821/commits/278a57863fe7a5895bc6cbf7657319096f8b62f3. Let me know if you think this helps/if you agree on the consistency since I've explained it. Another option would be change all three entries to "Key Pair" but I don't think that's necessary. Maybe being that explicit would be good though!

docs/appendices/glossary.rst
ChumOfChance: See comment above re: Journalist key

docs/appendices/glossary.rst
ChumOfChance: Good point! I improved the description and also remember the firewall, which is also under the admins purview! See https://github.com/freedomofpress/securedrop-docs/pull/821/commits/955b9011ff081247026f7aa1bed405a31e715bfe

docs/appendices/glossary.rst
ChumOfChance: In merging in the email alerts page, I also made a few edits there to reference the public/private alert keys explicitly as glossary terms where appropriate.

ChumOfChance:
Re: "Journalist Interface", I tried to remove all references to it, I think any left should be in sections that are pending a major re-write for a Qubes-only SDW. 

I think we should be removing references to it, because without Tails and without a Tor Browser configured with the authentication token, journalists will have no way to access it. As far as the plan is now, this web-based interface would only be accessed by administrators, so I think it makes sense to start pivoting the docs in this direction. If access to the older web-based Journalist Interface gets included on SDW for journalists as a fallback, then we'll write up new documentation for it that makes it clear it's a fallback. If newsrooms are still using Tails-based SecureDrop before it is EOL'd, they'll consult the older version of documentation. 

Previously there were a few glossarized *Admin Interface*s floating around the docs, so there is precedence for that term referring to the admin-only options of the JI, so that's the name I have gone with. In some cases I have tried to highlight that it is web-based and accessed via the Tor Browser, since that will be in stark contrast to the other ways users on the newsroom side interact with SecureDrop. The final name could be different, but maybe that will come when the exact implementation lands ("Admin Web GUI"?).

nathandyer:
Just a few thoughts on removing the italicization:

* Overall, I think I'm largely in favor of it (or at least, changing they way that we currently highlight glossary items)
* I think it's important that it be obvious when someone encounters a glossary term, whether it be capitalization, italicization, a link back to the glossary, etc. 
* The way we've traditionally done it is probably overkill, and IMO is a bit jarring sometimes to the point where it takes me out of what I'm reading

So with that in mind, I'm in favor of changing this up a bit, and keeping the capitalization but doing away with italicization seems like a good way to go about it.

I do think this is the kind of decision where broader team input is important.

docs/appendices/glossary.rst
nathandyer: Well said! I think my distinction was more that it sends alerts about events that are largely informational in nature, which _could_ point to a potential intrusion or other serious issue, and that really it's up to the Admins to use make the ultimate judgement calls. The change you made in https://github.com/freedomofpress/securedrop-docs/commit/5e5a25fa710992a716914a2d193ac6c73ccd2f48 captures that beautifully.

docs/appendices/glossary.rst
nathandyer: I think this change makes a lot of sense! I'm very happy with the main entry being "_____ Key" and referring to the _____ Public / Private keys in the descriptions (and throughout the docs) with Title Case. I agree that being explicit about when to use the public vs. private keys is very important, especially for particularly sensitive keys.

jskinne3 (dismissed):
Philosophical questions aside, this looks good!

Note for the future, here are some of those questions:

- Should we keep old (Tails) and new (Qubes) versions of the docs available?
- Should we keep some limited Tails stuff within the updated docs?
- Should we dream up a new name for the Admin Interface / Journalist Interface in its new, narrower role?

ChumOfChance:
I've just gone and resolved the many merge conflicts!

nathandyer (approved):
Everything here LGTM.

We met this morning to discuss the questions mentioned above, and decided that it makes sense to proceed with separate Qubes and Tails-based docs. With that in mind, we can move forward with accepting this as part of the docs re-write, and leave the original glossary in its current state elsewhere for the Tails docs.

Approving + merging.

PR #825, opened 2026-07-02
Document the process for v3 onion services key rotation

nathandyer:
This PR documents how to rotate keys for onion services, and apply them to the various workstations.

## Test plan
- [x] Ensure that SSH-over-Tor is already enabled
- [x] Follow the steps to rotate your onion service keys
- [x] Verify the keys change during the process
- [x] Verify you are able to regain SSH access by manually swapping the SSH-related keys
- [x] After rebooting, verify that you are able to access the Journalist interface
- [x] Verify you are still able to connect to both app and mon via SSH-over-Tor

## Checklist

This change accounts for:

### DIFF
--- a/docs/admin/deployment/offboarding.rst
+++ b/docs/admin/deployment/offboarding.rst
@@ -440,8 +440,8 @@ Then, move `tor_v3_keys.json` (which contains the keypairs for the *Journalist I
         *Journalist Interface*, and connect to both the *App* and *Mon* servers via
         SSH.
    
-      * For *SecureDrop Workstation*, repeat the import process as 
-        `outlined here. <https://workstation.securedrop.org/en/stable/admin/install/install.html#import-journalist-interface-details>`__
+      * For *SecureDrop Workstation*, manually copy the *Journalist Interface* details from the *Admin Workstation* to the configuration file in ``dom0`` and then apply the changes as
+        `outlined here. <https://workstation.securedrop.org/en/stable/admin/install/troubleshoot.html#failed-to-import-journalist-interface-details>`__
    
         Make sure to run:
    
--- a/docs/admin/deployment/offboarding.rst
+++ b/docs/admin/deployment/offboarding.rst
@@ -317,6 +317,133 @@ Return to the Secure Viewing Station
    and updating the fingerprint when prompted. You do not have to run
    ``securedrop-admin install`` again, since you have already pushed the
    changes to the server.
+   
+   
+Rotate Onion services keys
+--------------------------
+
+There are a three pairs of authenticated onion service keys that are used to
+secure access to various resources within a SecureDrop environment. These are:
+
+* Keys to access the Journalist interface
+* Keys to access the *Application Server* when SSH-over-Tor is enabled, and
+* Keys to access the *Monitor Server* when SSH-over-Tor is enabled
+
+To rotate these credentials, you can follow the steps below. 
+
+.. note::
+
+   Rotating the authenticated onion service keys will not change the Onion
+   addresses for the services. Your *Source Interface* and *Journalist
+   Interface* addresses will remain the same.
+
+#. Prepare your *Admin Workstation* to generate new keys: 
+   
+   .. code:: sh
+
+    rm ~/.config/securedrop-admin/app-journalist.auth_private
+    rm ~/.config/securedrop-admin/app-sourcev3.auth_private
+    mv ~/.config/securedrop-admin/tor_v3_keys.json ~/.config/securedrop-admin/tor_v3_keys.json.bak
+    
+#. Generate new keys on your *Admin Workstation*:
+
+   .. code:: sh
+   
+    securedrop-admin generate_v3_keys
+    
+#. Push the new keys to your SecureDrop server:
+
+   .. code:: sh
+   
+    securedrop-admin install
+    
+   .. important::
+
+      If you have SSH-over-Tor enabled, the first install run **will fail**
+      midway through the run, due to the keys for SSH getting swapped out. This
+      is expected. When you see this failure message, you need to do the following:
+   
+      1. Open the ``~/.config/securedrop-admin/tor_v3_keys.json`` file and locate the new private keys for both the *Application* and *Monitor Server* SSH access. They will be marked ``"app_ssh_private_key"`` and ``"mon_ssh_private_key"``.
+   
+      2. Open the ``~/.config/securedrop-admin/app-ssh.auth_private`` file, and replace the previous key (the last portion of the line, directly following ``:x25519:``) with the new key. 
+   
+      3. Repeat the process for ``~/.config/securedrop-admin/mon-ssh.auth_private``
+   
+      4. Reboot your *Admin Workstation*
+   
+      5. SSH into each server and confirm access
+   
+      6. Repeat the ``securedrop-admin install`` command, which should complete successfully this time
+    
+#. Configure your *Admin Workstation* to use the new keys:
+
+   .. code:: sh
+    
+    securedrop-admin localconfig
+    
+   and reboot when prompted.
+   
+#. Confirm you can login to the *Journalist Interface*
+
+#. Clean up leftover backup files
+
+   .. code:: sh
+   
+    rm ~/.config/securedrop-admin/tor_v3_keys.json.bak
+    rm ~/.config/securedrop-admin/app-journalist.auth_private.bak
+    rm ~/.config/securedrop-admin/app-sourcev3.auth_private.bak
+   
+#. Update the keys on *Journalist Workstation*, *SecureDrop Workstation*,
+   or other *Admin Workstation* drives.
+    
+      * For *Journalist Workstation* drives, copy:
+   
+        .. code:: sh
+   
+         ~/.config/securedrop-admin/app-journalist.auth_private
+    
+        from your *Admin Workstation* to a *Transfer Device*, then copy it
+        to the corresponding location on the *Journalist Workstation*. Apply
+        the changes via:
+    
+        .. code:: sh
+     
+         securedrop-admin localconfig
+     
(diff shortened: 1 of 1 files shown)


docs/admin/deployment/offboarding.rst
```diff
@@ -317,6 +317,133 @@ Return to the Secure Viewing Station
...
+--------------------------
+
+There are a three pairs of authenticated onion service keys that are used to
+secure access to various resources within a SecureDrop environment. These are:
+
+* Keys to access the Journalist interface
+* Keys to access the *Application Server* when SSH-over-Tor is enabled, and
+* Keys to access the *Monitor Server* when SSH-over-Tor is enabled
+
+To rotate these credentials, you can follow the steps below.
```
jskinne3: Set expectations here: will these instructions rotate all 3 pairs of keys at once? Or will they have to be repeated 3 times? Would a user ever want to rotate fewer than all 3?

docs/admin/deployment/offboarding.rst
```diff
@@ -317,6 +317,133 @@ Return to the Secure Viewing Station
...
+   Rotating the authenticated onion service keys will not change the Onion
+   addresses for the services. Your *Source Interface* and *Journalist
+   Interface* addresses will remain the same.
+
+#. Prepare your *Admin Workstation* to generate new keys: 
+   
+   .. code:: sh
+
+    rm ~/.config/securedrop-admin/app-journalist.auth_private
+    rm ~/.config/securedrop-admin/app-sourcev3.auth_private
```
ChumOfChance: Should this be `app-sourcev3-ths`?

ChumOfChance: Or should it be `app-ssh.auth_private`?

nathandyer: Good catch - it should be `app-sourcev3-ths`. You'll need to leave the SSH auth private keys in place at first.

docs/admin/deployment/offboarding.rst
```diff
@@ -317,6 +317,133 @@ Return to the Secure Viewing Station
...
+    mv ~/.config/securedrop-admin/tor_v3_keys.json ~/.config/securedrop-admin/tor_v3_keys.json.bak
```
ChumOfChance, line 357: It's easy to run through these steps quickly and miss that it's `rm`, `rm` and the `mv`. I think there is a good possibility someone who is not directly copy-pasting will end up `rm`ing their `tor_v3_keys.json` file. 

Instead, this section could be broken up and made more descriptive to explain to the user what's happening:

```suggestion
#. Prepare your *Admin Workstation* to generate new keys: 
   
First, delete the existing `app-journalist.auth_private` (containing the *Journalist Interface* private key and onion address) and `app-sourcev3.ths` (containing the *Source Interface* address):
   .. code:: sh

    rm ~/.config/securedrop-admin/app-journalist.auth_private
    rm ~/.config/securedrop-admin/app-sourcev3.ths
    
Then, move `tor_v3_keys.json` (which contains the keypairs for the *Journalist Interface* and the servers) to delete the original and leave a backup version of the file:

    .. code:: sh

    mv ~/.config/securedrop-admin/tor_v3_keys.json ~/.config/securedrop-admin/tor_v3_keys.json.bak
```

legoktm: We could have them open ~/.config/securedrop-admin with nautilus so they can do these rename/delete operations graphically?

ChumOfChance: Historically the rest of the documentation for administrators assumes they are comfortable with the commandline and provides the instructions as commands. I don't think we should break with that here, just the way the 3 commands were lined up it would be easy to miss and then accidentally run a destructive command. Myself reading the commands in a different context than usual (the GitHub review display) I almost made that mistake. 

And in general I think it's good to provide context for commands (not assuming the reader knows what the command is doing every time), so I've also added that here. It also slows the reader down so they notice the first 2 commands in this block are delete, and the third is move.

docs/admin/deployment/offboarding.rst
```diff
@@ -317,6 +317,133 @@ Return to the Secure Viewing Station
...
+
+   .. code:: sh
+   
+    securedrop-admin install
+    
+   .. important::
+
+      If you have SSH-over-Tor enabled, the first install run **will fail**
+      midway through the run, due to the keys for SSH getting swapped out. This
+      is expected. When you see this failure message, you need to do the following:
```
ChumOfChance: In my case the failure happens on this step:
```
RUNNING HANDLER [tor-hidden-services : Waiting for SSH connect (slow)...]
```

Proceeded by this step, which succeeds:
```
RUNNING HANDLER [tor-hidden-services : Clear out SSH connections, to prevent reuse of stale ControlPersist file.]
```

docs/admin/deployment/offboarding.rst
```diff
@@ -317,6 +317,133 @@ Return to the Secure Viewing Station
...
+   
+      6. Repeat the ``securedrop-admin install`` command, which should complete successfully this time
+    
+#. Configure your *Admin Workstation* to use the new keys:
+
+   .. code:: sh
+    
+    securedrop-admin localconfig
+    
+   and reboot when prompted.
```
ChumOfChance: I was not prompted to reboot at this step, but it's probably harmless to leave this instruction in?

docs/admin/deployment/offboarding.rst
nathandyer: Can confirm the exact same here, I'll document so admins will know precisely what to expect

docs/admin/deployment/offboarding.rst
```diff
@@ -317,6 +317,133 @@ Return to the Secure Viewing Station
...
+        .. code:: sh
+   
+         securedrop-admin localconfig
+    
+        and reboot when prompted. Once rebooted, verify that you can login to the
+        *Journalist Interface*, and connect to both the *App* and *Mon* servers via
+        SSH.
+   
+      * For *SecureDrop Workstation*, repeat the import process as 
+        `outlined here. <https://workstation.securedrop.org/en/stable/admin/install/install.html#import-journalist-interface-details>`__
```
ChumOfChance: That page advises running `sdw-admin --configure`, which will fail because there is already a valid configuration file.

Instead we could link to https://workstation.securedrop.org/en/stable/admin/install/troubleshoot.html#failed-to-import-journalist-interface-details and advise to follow the instructions just for importing the journalist key.  
```suggestion
      * For *SecureDrop Workstation*, manually copy the *Journalist Interface* details from the *Admin Workstation* to the configuration file in ``dom0`` and then apply the changes as
        `outlined here. <https://workstation.securedrop.org/en/stable/admin/install/troubleshoot.html#failed-to-import-journalist-interface-details>`__
```

This is still not ideal, I think jumping to that page it might be easy to get confused with the steps for the submission key. Another point for having dedicated built-in commands for rotating keys.

docs/admin/deployment/offboarding.rst
```diff
@@ -317,6 +317,133 @@ Return to the Secure Viewing Station
...
+   
+        .. code:: sh
+   
+         ~/.config/securedrop-admin/app-journalist.auth_private
+         ~/.config/securedrop-admin/app-ssh.auth_private
+         ~/.config/securedrop-admin/mon-ssh.auth_private
+         ~/.config/securedrop-admin/tor-v3-keys.json
+    
+        from the original *SecureDrop Workstation* to a *Transfer Device*, then copy
+        it to the corresponding location on the other *Admin Workstation.* Apply
```
ChumOfChance:
```suggestion
        them to the ``~/.config/securedrop-admin/`` directory on the other *Admin Workstation.* Apply
```

docs/admin/deployment/offboarding.rst
```diff
@@ -317,6 +317,133 @@ Return to the Secure Viewing Station
...
+   or other *Admin Workstation* drives.
+    
+      * For *Journalist Workstation* drives, copy:
+   
+        .. code:: sh
+   
+         ~/.config/securedrop-admin/app-journalist.auth_private
+    
+        from your *Admin Workstation* to a *Transfer Device*, then copy it
+        to the corresponding location on the *Journalist Workstation*. Apply
```
ChumOfChance:
```suggestion
        from your *Admin Workstation* to a *Transfer Device*, then copy it
        to the ``~/.config/securedrop-admin/`` directory on the *Journalist Workstation*. Apply
```

docs/admin/deployment/offboarding.rst
ChumOfChance: We should add this to the text, to flag which step we are expecting to fail on. If it doesn't fail here, then re-run the playbook.

docs/admin/deployment/offboarding.rst
nathandyer: I think so, although to be more accurate I'll adjust it to say "and reboot _if_ prompted"

docs/admin/deployment/offboarding.rst
nathandyer: Agreed. I think there's also an opportunity here to avoid a "choose your own adventure" step once Workstation docs are fully/publicly integrated into the main docs repo.

docs/admin/deployment/offboarding.rst
nathandyer: Thanks @jskinne3 - please let me know if the most recent revision better sets expectations.

(For what it's worth, I'm intentionally skating around the question of whether or not a user would ever want to rotate fewer than all 3. I can imagine a scenario where only the SSH keys or only the Journalist Interface key would want to swapped out, but right now our tooling doesn't allow for it.)

nathandyer:
@ChumOfChance Thanks for all the feedback and suggestions! This should be ready for another pass.

ChumOfChance:
I think this is ready to go!

legoktm (approved):
Per Martin

PR #830, opened 2026-07-21
Remove upgrade guide, and document the server and workstation update processes

### LINKED ISSUE #762: Consolodate upgrade guides into single upgrade guide

nathandyer:
Fixes #762

This PR combines the existing upgrade guides into a single upgrade guide, which highlights new features, migration paths, and other necessary upgrade details for users.

It covers server, workstation, and inbox, and provides at-a-glance information about the most recent releases.

**Note for reviewers:** I think it's better to focus on the overall format/outline, rather than the release-specific content (since the release details will change again between now and making these docs stable)

## Test plan
- [ ] Visual review

## Checklist

This change accounts for:


ChumOfChance (commented):
This is looking good! I especially like the summary table at the top. A few suggestions:

 - It would be useful to have a section at the top, explaining how the upgrades work for each piece:
   - SecureDrop servers are upgraded automatically, through unattended apt upgrades. You can verify an upgrade worked by checking the JI or SI.
   - The SecureDrop Workstation software is upgraded automatically as well, as part of pre-flight upgrades when you start your SecureDrop Workstation. This includes SecureDrop Inbox. Also note how to check if an upgrade has been applied/what version you are on (this seems to be a frequent question)
   - Also a good place to link to troubleshooting for update failures (in particular for Workstation/Inbox)
- I understand this page is just going to include information about the current releases, but it should also link to the respective changelogs.
- This page is no longer really an upgrade guide, so maybe the name should be changed. "Upgrading SecureDrop"?

legoktm:
So...I would prefer not to have the version numbers here, just as one less thing to update during the release process. We now have the Server + Workstation version numbers on the securedrop.org homepage, and we can add the Inbox there too.

I wonder how much we still need an upgrade guide vs just documentation on the Workstation updater? The server is automatic (+admin tools move to SDW's updater), Inbox doesn't get updated independently, so then it's just the Workstation itself. Am I missing something?

nathandyer:
If we can get away with not having an explicit upgrade guide at all, I'd definitely be into that!

The only place where users _need_ to know the version numbers is doing the git cloning process to bootstrap the admin tools on Tails, but with the bootstrapping no longer being necessary in the Qubes-ified version, I don't think that will be necessary at all.

Curious to hear @ChumOfChance's thoughts, but I'd be OK with changing this to a PR removing the (placeholder as it is anyway) upgrade guide page.

ChumOfChance:
I agree with removing an Upgrade Guides page. We should still include a page with the information about how the different parts of SecureDrop are upgraded, how to tell what version you are on, and how to troubleshoot a stuck update. This could all be on one page, or distributed on other pages (we already have troubleshooting for SDW, which is probably all we need for that). Current versions could be on the index page?

Is there an magic way to list the current versions on a docs page by pulling from GitHub or securedrop.org? Maybe an embedded GitHub version button?

Two user stories:

I am a SecureDrop Administrator, I have just heard about an update to SecureDrop and I want to check if my SecureDrop has successfully updated. I want to know where to look to find the version I'm on.

I am a SecureDrop Administrator. I have been somewhat absent from maintaining the project and I want to know the current versions of each SecureDrop component to compare to the versions I am running locally to confirm my SecureDrop has been updating correctly.

nathandyer:
Thanks @ChumOfChance!

> I agree with removing an Upgrade Guides page. We should still include a page with the information about how the different parts of SecureDrop are upgraded, how to tell what version you are on, and how to troubleshoot a stuck update. This could all be on one page, or distributed on other pages (we already have troubleshooting for SDW, which is probably all we need for that). Current versions could be on the index page?

Agreed! I just completely overhauled this branch and PR to do just that, in commit https://github.com/freedomofpress/securedrop-docs/commit/86e6e677bff27c4b223552684c0c2322276081fd

Please note that as part of this I've consolidated some of the existing update troubleshooting material into two distinct upgrade troubleshooting/info pages (one for server, one for Workstation/Inbox)

> Is there an magic way to list the current versions on a docs page by pulling from GitHub or securedrop.org? Maybe an embedded GitHub version button?

You can embed HTML here, so it should be possible. That's probably beyond the scope of this PR so for now I'm just linking out to securedrop.org.

docs/admin/workstation_reference/workstation_updates.rst
```diff
@@ -1,8 +1,29 @@
...
 After you unlock your SecureDrop Workstation, the preflight updater will prompt you to check for available
-system updates at least once per day.
+updates at least once per day.
+
+The preflight updater delivers new SecureDrop Inbox releases, SecureDrop Workstation updates, and updates to the underlying Qubes operating system.
+
+Verifying version numbers
+-------------------------
+
+`The current SecureDrop Inbox and Workstation version numbers are available at securedrop.org. <https://securedrop.org/>`__
```
ChumOfChance: "Latest" makes a better distinction between what the user is (*current*ly) running and the most recent release. It's also the language that appears on the linked page. 

```suggestion
`The latest SecureDrop Inbox and Workstation version numbers are available at securedrop.org. <https://securedrop.org/>`__
```

docs/admin/workstation_reference/workstation_updates.rst
```diff
@@ -1,8 +1,29 @@
...
+
+You can find the version number of the SecureDrop Inbox running on your workstation on the login screen when you open the Inbox.
+
+The version number for the SecureDrop Workstation itself can be found by running: ::
+  
+  sudo dnf list securedrop-workstation-dom0-config
+  
+in a ``dom0`` Terminal.
```
ChumOfChance: Terminal before the command to run.

```suggestion
You can find the version number of the SecureDrop Inbox running on your workstation on the login screen when you open the Inbox.

The version number for the SecureDrop Workstation itself can be found by running the following command in a ``dom0`` Terminal: ::
  
  sudo dnf list securedrop-workstation-dom0-config
```

docs/admin/workstation_reference/workstation_updates.rst
ChumOfChance:
```suggestion
Below are some troubleshooting steps for common
```

docs/admin/maintenance/server_updates.rst
```diff
@@ -1,5 +1,70 @@
+Updating SecureDrop server
+==========================
+
+Automatic updates
+-----------------
+
+SecureDrop servers check for updates automatically, and install updates without any manual action or intervention within 24 hours of each new SecureDrop release.
+
+`The current SecureDrop server version number is available at securedrop.org. <https://securedrop.org/>`__
```
ChumOfChance:
```suggestion
`The latest SecureDrop server version number is available at securedrop.org. <https://securedrop.org/>`__
```

docs/admin/maintenance/server_updates.rst
```diff
@@ -1,5 +1,70 @@
...
+  sudo apt update
+  sudo apt upgrade
+
+from both the Application and Monitor servers.
+
+You can verify the installed version number by running: ::
+
+    sudo apt policy securedrop-app-code
+    
+The current SecureDrop version is also visible within the Source Interface.
```
ChumOfChance:
```suggestion
The current SecureDrop version is also visible at the bottom of the Source Interface.
```

ChumOfChance (changes requested):
I do think adding GitHub version badges would be appropriate in this PR, it feels a bit less clunky than linking out to another page. But it could be saved for later work. I've made a few suggestions.

nathandyer:
I've introduced a change which now dynamically shows the latest version of the various SecureDrop components (by way of the latest tag). It _does_ introduce a page load from shields.io, which may not be desirable. If there's a better method that doesn't require accessing a third-party resource, I'm certainly open to it.

ChumOfChance:
I guess a third party asset (besides the necessary GitHub) in the docs isn't great. It looks like a solution would be to host our own shields server: https://hub.docker.com/r/shieldsio/shields/ which would be quite cool, but that certainly is a topic for a bigger PR. So maybe your initial instinct was right, we should just include a link for now. Sorry for the wild goose chase. It does look cool!

nathandyer:
Now that I've seen it I really _want_ it to work that way!

But for now at least, I've dropped the latest commit, and we can follow up with self-hosting options a little bit further down the road.

ChumOfChance (approved):
All looks good to me now!

PR #832, opened 2026-07-27
Remove hard line wrapping and enforce one sentence per line

### LINKED ISSUE #815: Remove hard line wrapping
## Describe the change
Our [Documentation Guidelines](https://developers.securedrop.org/en/latest/documentation_guidelines.html#line-wrapping) specify hard wrapping the documentation source at 80 characters, with some exceptions. This is no longer applied consistently, and I am not aware of any rationale for keeping it. The only documented reference I can find to this policy is here: https://github.com/freedomofpress/securedrop-docs/pull/367#issuecomment-1204747495 which suggests that even 3 years ago it was unclear why we recommend this.

Besides (imo reduced) readability of the source, the hard wrap can cause problems when searching for terms or doing a find/replace when a line break occurs in the middle of a multi-word term (depending on the editor used).

I would like to remove all the line breaks that enforce the 80 character line wrap and going forward change the published guidelines to remove this policy.

Making this change would be somewhat destructive, it would be a pretty big diff despite just removing line breaks. There might be other reasons against doing this that I am not aware of!

nathandyer:
Fixes #815 

This PR:
- [ ] Removes hard line wrapping
- [ ] Moves every sentence to its own line
- [ ] Adds an extra empty line between paragraphs
- [ ] Creates the `.git-blame-ignore-revs` file and marks the commit making the above changes as ignored

Note: In #815 @ChumOfChance mentioned we might want to take the conservative approach of _not_ putting each sentence on its own line. Doing so was about the same amount of work either way, so I opted to do so (if we think we might ever _want_ make a big change, might as well do it now)

If we decide that is _not_ ideal, I can modify this to eliminate separate lines for each sentence.

Also of note: I don't know that I _love_ having multiple empty lines between paragraphs. Curious to hear others' thoughts. If we want to change that, I can automatically run through it and eliminate the extra empty lines.

## Test plan
- [ ] CI is happy
- [ ] Previewing locally looks OK
- [ ] No issues with overall "look and feel" of the RST files

## Checklist

This change accounts for:


docs/admin/deployment/ssh_over_local_net.rst
```diff
@@ -46,50 +49,59 @@ Configuring SSH for local access
...
 
-    Username for SSH access to the servers: vagrant
-    Local IPv4 address for the Application Server: 10.0.1.4
-    Local IPv4 address for the Monitor Server: 10.0.1.5
-    Hostname for Application Server: app
-    Hostname for Monitor Server: mon
-    [...]
-    Enable SSH over Tor (recommended, disables SSH over LAN). If you respond no, SSH will be available over LAN only: no
+
+    Username for SSH access to the servers: vagrant Local IPv4 address for the Application Server: 10.0.1.4 Local IPv4 address for the Monitor Server: 10.0.1.5 Hostname for Application Server: app Hostname for Monitor Server: mon [...] Enable SSH over Tor (recommended, disables SSH over LAN).
```
nathandyer, line 73: Looks like this is busted, we should pay extra attention to anywhere there's a list

cfm:
I think I like this!  However, I imagine the real test will come with changes *on top of* this change, to see what the one-sentence-per-line diff looks like.


> I don't know that I _love_ having multiple empty lines between paragraphs. Curious to hear others' thoughts. If we want to change that, I can automatically run through it and eliminate the extra empty lines.

Yes, if the original suggestion in #815 was one line break between sentences, two between paragraphs, then I think this is _three_ between paragraphs?

ChumOfChance:
Yeah, I was also just looking at this.  I have not done a close look, but my initial bird's eye take is:

- I think there was a misunderstanding on the line breaks, this PR currently adds an extra blank line between paragraphs that I think no one wants. "Two line breaks" was meant to include the line break at the end of the last line of the previous paragraph, followed by an additional line break (yielding one blank line).
- I think I prefer *not* having one sentence = one line. To me, it's nicer for the source to look more like the final rendered text. Perhaps we would just get used to, but I also think the change would make it harder for outside contributors to get it right since it's a bit unusual. Unless someone feels very strongly about it?

nathandyer:
Thanks both, that's very helpful!

I started to make changes to the foundation here, but I'm realizing there are a lot of issues with the current state of the PR after my pass yesterday. 

I think I'll close this PR and start fresh with that slightly new framing in mind.

Issue #817, opened 2026-06-18
Update instructions for provisioning export devices for post-Tails


ChumOfChance:
The current documentation for provisioning encrypted USB flash drives for exporting transcripts and attachments ("Export Devices") are for Tails. New instructions are needed to provision export devices on Qubes for future SecureDrops using only Qubes-based SecureDrop Workstations.

PR #833, opened 2026-07-27
Update Export Device secure erasure instructions and recommendations

### LINKED ISSUE #563: Review and update recommendations regarding secure deletion
## Describe the change
We should ensure our recommendations regarding secure deletion make sense for USB/SSD deletion, and that we help journalists mitigate risks with files that could be recovered by an adversary who gains access to a drive and passphrase.

This is especially true for the export device, which holds files in decrypted form (the drive itself is meant to be encrypted; our docs recommend VeraCrypt for this device, as it may be opened on non-Linux machines).

For example, we could recommend a regular drive-level overwrite operation for folks who routinely re-use a single export device -- but we'd need to ensure that such a recommendation actually meaningfully mitigates risk.

## How will this impact [SecureDrop users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?

- mitigate risk from search & seizures

## Additional context

- Tails 6 removed support for ["secure wipe"](https://tails.net/doc/encryption_and_privacy/secure_deletion/) functionality, since the reliability of such wipe operations on a per-file basis cannot be guaranteed on USB/SSD devices.
- #558 updated the docs accordingly and no longer recommends "securely" wiping individual files

ChumOfChance:
Fixes #563

Based on the discussion in #563, this PR updates the recommendation for how to securely erase Export Devices.

- Update the instructions for decommissioning SecureDrop to reflect that VeraCrypt is used as well, remove references to the SVS and Tails, and direct readers to the instruction page for creating Export Devices. 
- Add a note on securely erasing Export Devices with a recommendation to do so regularly to the provisioning Export Devices page. 
- Add a similar note on securely erasing Export Devices with a recommendation to do so regularly the Journalist documentation page for working with submissions/exporting submissions. Journalists should also know how and why to securely delete files from Export Devices.

Note that future work is required on the instruction page for creating Export Devices (see https://github.com/freedomofpress/securedrop-docs/issues/817)

## Checklist

This change accounts for:

### DIFF
--- a/docs/admin/installation/provisioning_usb.rst
+++ b/docs/admin/installation/provisioning_usb.rst
@@ -88,3 +88,12 @@ Creating a VeraCrypt-encrypted drive
   its decryption password to the Journalists who will be using it. Make sure that
   they store it and its password securely, as it will contain decrypted
   submissions.
+
+Securely erase Export Devices
+-----------------------------
+
+Even if a Journalist regularly deletes files from an Export Device, there remains the possibility that those files will be recoverable by anyone with the decryption password. Because of the underlying technology, simply overwriting data or other "file shredding" techniques are insufficient to securely delete files from USB flash drives.
+
+To protect against the possibility that a compromised Export Device decryption password will give someone access to traces or the full contents of previously exported files, you should securely erase your Export Devices on a regular basis. This is done by re-formatting and re-encrypting the USB flash drive with a new encryption password, following the steps above for a LUKS or VeraCrypt drive as appropriate. You may then reissue the Export Devices to Journalists and provide them with the new decryption password.
+
+You may also choose to destroy the drives by physical means, such as using a hammer or purpose-built shredder to pulverize the drive.
--- a/docs/admin/installation/provisioning_usb.rst
+++ b/docs/admin/installation/provisioning_usb.rst
@@ -92,8 +92,8 @@ Creating a VeraCrypt-encrypted drive
 Securely erase Export Devices
 -----------------------------
 
-Even if a Journalist regularly deletes files from an Export Device, there remains the possibility that those files will be recoverable by anyone with the decryption password. Because of the underlying technology, simply overwriting data or other "file shredding" techniques are insufficient to securely delete files from USB flash drives.
+Due to the underlying technology of USB flash drives, Export Devices are likely to retain recoverable portions of files that have been stored on them even if those files have been delete. 
 
-To protect against the possibility that a compromised Export Device decryption password will give someone access to traces or the full contents of previously exported files, you should securely erase your Export Devices on a regular basis. This is done by re-formatting and re-encrypting the USB flash drive with a new encryption password, following the steps above for a LUKS or VeraCrypt drive as appropriate. You may then reissue the Export Devices to Journalists and provide them with the new decryption password.
+To protect against the possibility that a compromised encryption password will yield access to traces of previously exported files, you should securely erase your Export Devices on a regular basis. This can only be done by re-formatting and re-encrypting the USB flash drive with a new encryption password, following the steps above for a LUKS or VeraCrypt drive as appropriate. You may then reissue the Export Devices to Journalists and provide them with the new decryption password.
 
 You may also choose to destroy the drives by physical means, such as using a hammer or purpose-built shredder to pulverize the drive.
--- a/docs/admin/installation/provisioning_usb.rst
+++ b/docs/admin/installation/provisioning_usb.rst
@@ -92,7 +92,7 @@ Creating a VeraCrypt-encrypted drive
 Securely erase Export Devices
 -----------------------------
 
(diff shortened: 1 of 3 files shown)


docs/admin/installation/provisioning_usb.rst
```diff
@@ -88,3 +88,12 @@ Creating a VeraCrypt-encrypted drive
   its decryption password to the Journalists who will be using it. Make sure that
   they store it and its password securely, as it will contain decrypted
   submissions.
+
+Securely erase Export Devices
+-----------------------------
+
+Due to the underlying technology of USB flash drives, Export Devices are likely to retain recoverable portions of files that have been stored on them even if those files have been delete.
```
nathandyer:
```suggestion
Due to the underlying technology of USB flash drives, Export Devices are likely to retain recoverable portions of files that have been stored on them even if those files have been deleted. 
```

docs/admin/installation/provisioning_usb.rst
```diff
@@ -88,3 +88,12 @@ Creating a VeraCrypt-encrypted drive
...
   submissions.
+
+Securely erase Export Devices
+-----------------------------
+
+Due to the underlying technology of USB flash drives, Export Devices are likely to retain recoverable portions of files that have been stored on them even if those files have been delete. 
+
+To protect against the possibility that a compromised encryption password will yield access to traces of previously exported files, you should securely erase your Export Devices on a regular basis. This can only be done by re-formatting and re-encrypting the USB flash drive with a new encryption password, following the steps above for a LUKS or VeraCrypt drive as appropriate. You may then reissue the Export Devices to Journalists and provide them with the new decryption password.
+
+You may also choose to destroy the drives by physical means, such as using a hammer or purpose-built shredder to pulverize the drive.
```
nathandyer, line 99: Should we provide guidance on _when_ to destroy the drives? Presumably it's for when a drive is no longer needed, but are there scenarios where that would be preferred over re-encrypting with a new passphrase (even if an org wanted to keep reusing it)?

docs/journalist/submissions.rst
```diff
@@ -151,6 +148,16 @@ audio, and begin publishing important, high-impact work!
          </admin/deployment/getting_the_most_out_of_securedrop>` to read
          about encouraging sources to use SecureDrop.
 
+Securely erase an Export Device
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Due to the underlying technology of USB flash drives, Export Devices are likely to retain recoverable portions of files that have been stored on them even if those files have been delete.
```
nathandyer:
```suggestion
Due to the underlying technology of USB flash drives, Export Devices are likely to retain recoverable portions of files that have been stored on them even if those files have been deleted. 
```

docs/journalist/submissions.rst
```diff
@@ -151,6 +148,16 @@ audio, and begin publishing important, high-impact work!
...
+Securely erase an Export Device
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Due to the underlying technology of USB flash drives, Export Devices are likely to retain recoverable portions of files that have been stored on them even if those files have been delete. 
+
+To protect against the possibility that a compromised encryption password will yield access to traces of previously exported files, you should securely erase your Export Devices on a regular basis. You may also wish to do this on a case-by-case basis after handling particularly sensitive files or if you believe a decryption password may be compromised. 
+
+Securely erasing an Export Device can only be done by re-formatting and re-encrypting the USB flash drive with a new encryption password. You can follow the :doc:`same steps</admin/installation/provisioning_usb>` used to initially create the Export Device, or contact your Administrator.
+
+You may also choose to destroy the drives by physical means, such as using a hammer or purpose-built shredder to pulverize the drive.
```
nathandyer, line 160: Just flagging same question as above

nathandyer (changes requested):
Overall, this is looking great!

One small typo (in two places), and one question (also applicable in two places)

docs/admin/installation/provisioning_usb.rst
ChumOfChance: I think doing do could get into the territory of providing a lot of speculative advice. It's really just a more extreme option at the user's discretion. It could be a scenario where they absolutely 100% want to ensure the data is totally not recoverable, such as in decommissioning or as you say when a drive is no longer needed. Or the institutions internal policies could dictated that media is physically destroyed in this way (common for various compliances/certifications), and so mentioning that it is also an acceptable option means that our recommendation doesn't have to conflict with internal policy. Finally there's the tinfoil hat scenario, where the user is in a crisis situation, doesn't have a computer that can be trusted, needs the thing made unrecoverable now. 

But I think enumerating those options is a bit out of scope? If you think it would improve the flow, we could add "At your discretion, or if mandated by your organizations policies, you may also choose..." ?

nathandyer: I think you're right, and I agree that enumerating those options is a bit out of scope. Let's stick with how it is for now.

nathandyer (approved):
LGTM!

PR #835, opened 2026-07-28
Implement Sphinx Book theme

### LINKED ISSUE #834: Enable Sphinx "Book" theme
After a successful [trial in the dev docs](https://github.com/freedomofpress/securedrop-dev-docs/pull/300), let's go ahead and enable the new Book theme here as well (in `main` until we're ready to push all the new docs re-write work to stable)

ChumOfChance:
Fixes #834

This PR brings the [book sphinx theme](https://sphinx-book-theme.readthedocs.io/) to the SecureDrop documentation

This theme is somewhat more strict (or at least has different interpretations) in rendering rst compared to the default Read The Docs theme we were using. As such I also had to go through and fix a lot of extra indentations, lists (especially nested ones), and code blocks. I also found it was necessary to remove some image formatting we had before. While there, I made some small drive-by changes, mostly to match the preferred rst style (since that seems to be more important with this theme) or improve source readability.

I also undid https://github.com/freedomofpress/securedrop-docs/pull/734, assuming we no longer need the 2.12 build (and there is no longer a version picker to easily access it).

Otherwise, this is the stock book theme, with the addition of a download link for the rendered PDF (at the bottom of the left menu), some small CSS tweaks to colours in dark mode, and a few configuration options to match the features of our current docs.

## Test plan
- [ ] visual review, especially check for odd formatting problems


nathandyer (approved):
This all looks great! I'd love to nitpick it but I think it's good to land as-is :)

> This theme is somewhat more strict (or at least has different interpretations) in rendering rst compared to the default Read The Docs theme we were using. As such I also had to go through and fix a lot of extra indentations, lists (especially nested ones), and code blocks. I also found it was necessary to remove some image formatting we had before. While there, I made some small drive-by changes, mostly to match the preferred rst style (since that seems to be more important with this theme) or improve source readability.

These changes all LGTM, and were sorely needed! I think this will pay dividends in other areas where strict RST formatting is necessary.

> I also undid #734, assuming we no longer need the 2.12 build (and there is no longer a version picker to easily access it).

That seems reasonable to me, I don't think we'll need it post-rewrite launch. And in line with the discussion from last week, we're OK to not have a direct switcher to the latest builds (but they'll still be accessible from the URL if needed by devs)

> Otherwise, this is the stock book theme, with the addition of a download link for the rendered PDF (at the bottom of the left menu), some small CSS tweaks to colours in dark mode, and a few configuration options to match the features of our current docs.

## Test plan 
* [x]  visual review, especially check for odd formatting problems

deeplow:
Just dropping here what this change looks like, for future reference:

| Before | After |
|--|--|
| Before | After |

PR #839, opened 2026-07-29
Remove unused images and includes

### LINKED ISSUE #819: Remove unused include files and dead references
There are several `.txt` includes in `/docs/includes/`. `squash-commits.txt` and `update-gui.txt` are examples. I think these vestigial files can be deleted.

Similarly there are plenty of dead targets for `:ref:` links. There may be an automated way to detect and remove these dead links.

ChumOfChance:
Fixes #819

This PR removes includes and images that are no longer used. I left in the diagrams folder, and the complete set of screenshots for the Journalist and Source Interfaces, since these still may be useful as a reference.

Also removed are any URLs that were included not-inline (below a paragraph) but for which the original link had been since deleted. 

## Test plan
- [ ] CI happy
- [ ] Nothing removed will be missed

## Checklist

This change accounts for:


legoktm:
Wow. Feels like we would benefit from some sort of CI check to find these kinds of unused resources.

ChumOfChance:
@legoktm possibly, I didn't note it in this draft yet but I did leave several images in that are no longer used. There is a full collection of diagrams (included outdated ones) and complete screenshots for the JI and SI. Even though not every file is referenced, it seems useful to keep them on-hand in the repo as long as they are still relevant.

legoktm:
> There is a full collection of diagrams (included outdated ones) and complete screenshots for the JI and SI. Even though not every file is referenced, it seems useful to keep them on-hand in the repo as long as they are still relevant.

Maybe we could have a dedicated archive/ folder or something for those types of files? Or some other way of intentionally marking them as unused.

ChumOfChance:
@legoktm I think there is a good case for a reorganization of the images, which warrants another PR. Currently the only extra images (I believe) are sorted in intuitive locations, just the ones mentioned above: a folder of diagrams (with a readme!) and a folder of complete screenshots of the SI and JI (not all of which are used). 

I think the best time to do this will be in the final stages after bringing in all the sd-admin stuff, as there are also some final integrations of the SDW docs steps to do (like the `workstation_reference` folder).

legoktm:
Sounds good to me, ty!! :)

nathandyer (approved):
LGTM! The linter (both CI and an additional local check on my side) is still happy, so it doesn't appear that anything that was removed here was still in use.

PR #850, opened 2026-09-08
Update landing page and source guides with LLM recommendations

### LINKED ISSUE #849: Add a new landing page guideline to warn against using LLMs to aid with SecureDrop submissions
We should add additional guidance to our landing page documentation, recommending that newsrooms adopt language to warn sources against submitting confidential information to Claude, ChatGPT, etc. prior to submitting to a SecureDrop.

zenmonkeykstop:
Fixes #849 

## Test plan
Changes are consistent with those discussed in #849, and read well.

## Checklist

This change accounts for:


docs/admin/deployment/landing_page.rst
```diff
@@ -49,7 +49,7 @@ You can use a separate computer you've designated specifically to handle the sub
...
 
-Your online behavior can be extremely revealing. Regularly monitoring our publication's social media or website can potentially flag you as a source. Take great care to think about what your online behavior might reveal, and consider using Tor Browser to mitigate such monitoring.
+Your online behavior can be extremely revealing. Regularly monitoring our publication's social media or website can potentially flag you as a source. Your activity on other websites or apps may also be used to identify you. Avoid using any service linked to your identity while preparing to contact us or while whistleblowing, such as LLM chatbots or search engines when logged in, as your prompt and search histories may be used against you. Take great care to think about what your online behaviour might reveal. If possible, use Tor Browser for all whistleblowing-related activity to protect your anonymity.
```
nathandyer: Nit: I don't know if this is officially part of our guidelines, but we seem to use en_US throughout the docs, and have at least four other instances where we refer to "behavior" rather than "behaviour." I think we should aim for consistency.

One other note: even though LLM is the more technically correct term, in case sources are either unfamiliar with the term or are searching specifically for guidance on AI, I think we need to at least refer to AI in the text.

```suggestion
Your online behavior can be extremely revealing. Regularly monitoring our publication's social media or website can potentially flag you as a source. Your activity on other websites or apps may also be used to identify you. Avoid using any service linked to your identity while preparing to contact us or while whistleblowing, such as LLM (AI) chatbots or search engines when logged in, as your prompt and search histories may be used against you. Take great care to think about what your online behavior might reveal. If possible, use Tor Browser for all whistleblowing-related activity to protect your anonymity.
```

docs/source/before_you_submit.rst
```diff
@@ -4,6 +4,7 @@ Before you submit
 What NOT to do
 --------------
 
+* DO NOT use online services that are linked to your identity (such as LLM chatbots) to look for whistleblowing resources.
```
nathandyer: Similar to above suggestion:

```suggestion
* DO NOT use online services that are linked to your identity (such as LLM/AI chatbots) to look for whistleblowing resources.
```

docs/source/before_you_submit.rst
```diff
@@ -50,7 +51,7 @@ While using Tor Browser on your personal computer helps hide your activity on th
 Choose who to submit to
 -----------------------
 
-We recommend conducting all research related to your submission in Tor Browser. If you are unsure whether you are using Tor, you can visit the address https://check.torproject.org.
+We recommend conducting all research related to your submission in Tor Browser. If you are unsure whether you are using Tor, you can visit the address https://check.torproject.org. Tor Browser provides strong anonymity protection, but you should also take care to ensure your research is not linked to your identity. Avoid the use of services that require you to provide persononally identifying information to use them, such as LLM chatbots, or search engines while logged in.
```
nathandyer: IMO, if we refer to AI elsewhere (above), then I think we're okay to omit it here. But happy either way.

nathandyer (changes requested):
One incredibly pedantic nit, and a general suggestion to explicitly use the term "AI", but otherwise this is looking great!

docs/admin/deployment/landing_page.rst
zenmonkeykstop: Aye to both, we can leave dust-ups over spelling to our respective national leaders :)

nathandyer (dismissed):
LGTM! Marking an approval, but waiting to merge in case others want to provide further input or suggestions

legoktm:
I don't know if editorial has guidance on this, but I would solely use the term AI and not LLM at all, which I think for most people is jargon; and the people who know what "LLM" means will also know that "AI" is referring to that.

legoktm (changes requested):
Per above.

zenmonkeykstop:
Updated to use "AI chatbots" instead, editorial concurs.

legoktm (approved):
Thanks, that sounds good to me. Will leave final review + approval for Nathan

nathandyer (approved):
LGTM, also a big fan of the change to "AI Chatbots"

